PR #25, opened 2015-01-25
Track line and position-within-line numbers

mcclure:
"Sedlexing now tracks position by a position_info structure instead of int. In addition to array position, position_info tracks line and position in line of each point in the buffer (where "line" is demarcated by `\n`)"

Things I am uncertain about here:
- I added two "FIXME: Requesting comment"s in sedlexing.ml for lines that I didn't understand.
- Is it sufficient to use `\n` as the sole determinor of line/linepos? (You won't work right on Mac OS 9 systems!)
- Should any other position information (such as byte offset) be tracked?
- Should there be any way to customize the kind of position information tracked?
- In order to keep the old API stable, I added the accessors in sedlexing.mli at "lexeme_start_extended", "lexeme_end_extended". Are these good names?

Here's my test code for this change:

```
  (* Simple sedlex sample *)
  let () = 
        let position_info i = "[pos:"^(string_of_int i.Sedlexing.position)^
              " line:"^(string_of_int i.Sedlexing.line)^
              " inline:"^(string_of_int i.Sedlexing.line_position)^"]" in
        let rec lex buf = 
              match%sedlex buf with
                    | eof -> print_endline "\tEnd"
                    | any ->
                          print_endline @@ "("^(Sedlexing.Utf8.lexeme buf)^")"^
                                " start " ^(position_info @@ Sedlexing.lexeme_start_extended buf)^
                                " end "   ^(position_info @@ Sedlexing.lexeme_end_extended   buf);
                          lex buf
                    | _ -> failwith "Internal failure: Reached impossible place"
        in lex @@ Sedlexing.Utf8.from_string "a   \nbc\n\n    \n"
```

### DIFF
--- a/src/lib/sedlexing.mli
+++ b/src/lib/sedlexing.mli
@@ -22,6 +22,17 @@
     Of course, you'll probably want to define functions like [lexeme] to
     be used in the lexers semantic actions.  *)
 
+type position_info = {
+  position : int;
+  line : int;
+  line_position : int;
+}
+      (** Lexer position info structure. Fields:
+          * position: offset in stream
+          * line: number of newlines preceding offset
+          * line_position: offset since beginning of line
+          position, line, and line_position all start at 0 *)
+
 type lexbuf
       (** The type of lexer buffers. A lexer buffer is the argument passed
           to the scanning functions defined by the generated lexers.
@@ -75,6 +86,19 @@ val loc: lexbuf -> int * int
     (** [Sedlexing.loc lexbuf] returns the pair
         [(Sedlexing.lexeme_start lexbuf,Sedlexing.lexeme_end
         lexbuf)]. *)
+
+val lexeme_start_extended: lexbuf -> position_info
+    (** [Sedlexing.lexeme_start_extended lexbuf] returns a position info
+        structure for the first code point of the matched string. *)
+
+val lexeme_end_extended: lexbuf -> position_info
+    (** [Sedlexing.lexeme_end_extended lexbuf] returns a position info
+        structure for the last code point of the matched string. *)
+
+val loc_extended: lexbuf -> position_info * position_info
+    (** [Sedlexing.loc_extended lexbuf] returns the pair
+        [(Sedlexing.lexeme_start_extended lexbuf,
+        Sedlexing.lexeme_end_extended lexbuf)]. *)
 
 val lexeme_length: lexbuf -> int
     (** [Sedlexing.loc lexbuf] returns the difference
--- a/src/lib/sedlexing.mli
+++ b/src/lib/sedlexing.mli
@@ -22,17 +22,6 @@
     Of course, you'll probably want to define functions like [lexeme] to
     be used in the lexers semantic actions.  *)
 
-type position_info = {
-  position : int;
-  line : int;
-  line_position : int;
-}
-      (** Lexer position info structure. Fields:
-          * position: offset in stream
-          * line: number of newlines preceding offset
-          * line_position: offset since beginning of line
-          position, line, and line_position all start at 0 *)
-
 type lexbuf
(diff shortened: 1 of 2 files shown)


src/lib/sedlexing.mli
```diff
@@ -22,6 +22,17 @@
...
+  position : int;
+  line : int;
+  line_position : int;
+}
+      (** Lexer position info structure. Fields:
+          * position: offset in stream
+          * line: number of newlines preceding offset
+          * line_position: offset since beginning of line
+          position, line, and line_position all start at 0 *)
+
```
Drup: I think it would be better to just use `Lexing.position`. It's the de-facto standard, since the menhir interface is using it.

We could take an optional argument when creating the lexbuf for the file name and use a dummy one otherwise.

src/lib/sedlexing.mli
```diff
@@ -76,6 +87,19 @@ val loc: lexbuf -> int * int
...
+val lexeme_end_extended: lexbuf -> position_info
+    (** [Sedlexing.lexeme_end_extended lexbuf] returns a position info
+        structure for the last code point of the matched string. *)
+
+val loc_extended: lexbuf -> position_info * position_info
+    (** [Sedlexing.loc_extended lexbuf] returns the pair
+        [(Sedlexing.lexeme_start_extended lexbuf,
+        Sedlexing.lexeme_end_extended lexbuf)]. *)
+
```
Drup: I would replace "extended" by "position". I think it's clearer.

mcclure: That makes sense, are you sure `loc_position` is not weird though?

Drup: it's a bit weird indeed, but I don't have a better consistent idea.

Drup:
So, in this proposal, the newline counting is automatic and the user can't affect it. It's quite different from ocamlex's solution where the user count the lines himself (with `Lexing.new_line`).

Are there reasons that will make the second preferable ? I'm not sure.

Drup:
In particular, with respect to UTF8 and the newline char.

mcclure:
So, if it is not automatic, I'm not sure why it's helpful to have it in Sedlex at all-- if I am already modifying my pattern to differentiate `\n` from other white_space, I can pretty easily keep track of buffer position -> line position mapping myself. The reason I am writing this patch is because I want to be able to simplify my patterns in order to not have to necessarily treat `\n` special anymore (see `quotedString` in <https://bitbucket.org/runhello/emily/src/fb5e47b51e8a60597589ce711f7c84ec9863a9eb/src/tokenize.ml?at=default> for example-- `stateNewline` is my local equivalent of `Lexing.new_line`) and for the peace of mind that I won't accidentally FORGET in one of my patterns to treat \n special and thus ruin the bookkeeping for the entire rest of the file.

If I'm right that what's desirable here is for the tracking to be "automatic", then if we wanted to allow customizing things, I think the best way to do it would be for the user to somehow provide some kind of instruction (a character set, a function, a regexp? I guess the options for specifying CRLF get awkward quick) that would be used to determine where the logical newline is located.

mcclure:
Addressed issues above *other* than the customizability / new_line thing. I guess really this shouldn't be merged in unless we support UTF-8 sensibly.

Drup:
> So, if it is not automatic, I'm not sure why it's helpful to have it in Sedlex at all

It's very helpful because: 
 - It removes boilerplate in the user code
 - It allows to plug sedlex and menhir almost directly.

I agree automatic is better if it can be pulled of but it's not that easy and you can rest assured that someone, somewhere, is going to want to count manually. :)

Drup:
I think the recent set of PR pretty much covers all the functionalities. Better late than never! Closing.

PR #29, opened 2015-03-13
Rename Cset -> Sedlex_cset

### LINKED ISSUE #28: Prefix Cset with something
This is a common enough name that it conflicts with ocaml-re for example.

rgrinberg:
Fix #28


alainfrisch:
Thanks!

PR #30, opened 2015-04-09
Make sedlexing.cma a separate findlib package

### LINKED ISSUE #22: ppx in META prevents using sedlex in libraries
In particular, if I distribute a library that depends on Sedlexing, adding `requires = "sedlex"` will also pull in `-ppx ppx_sedlex`, which is not what I want and will of course fail if the client is using camlp4.

whitequark:
In this way, a library depending on sedlex can use -package sedlex during
compilation, and requires = "sedlex.lib" in META file.

Fixes #22.

### DIFF
--- a/META
+++ b/META
@@ -1,7 +1,6 @@
 version = "1.99.2"
 description = "Runtime support for sedlex"
-archive(byte) = "sedlexing.cma"
-archive(native) = "sedlexing.cmxa"
+requires = "sedlex.runtime"
 ppx = "./ppx_sedlex"
 
 package "ppx" (
@@ -10,4 +9,11 @@ package "ppx" (
   requires = "ppx_tools gen"
   archive(byte) = "sedlex.cma"
   archive(native) = "sedlex.cmxa"
+)
+
+package "runtime" (
+  version = "1.99.2"
+  description = "runtime support library of sedlex"
+  archive(byte) = "sedlexing.cma"
+  archive(native) = "sedlexing.cmxa"
 )


darioteixeira:
@whitequark: I assume you're following the discussion on the caml-list, and it seems you're opting for the alternative scheme. Do you have strong arguments to favour it over the original scheme?

whitequark:
I have no idea about any discussion on caml-list, neither I really think one is needed. I did the simplest possible thing while keeping backwards compatibility.

darioteixeira:
@whitequark: The discussion thread [starts here](https://sympa.inria.fr/sympa/arc/caml-list/2015-04/msg00036.html), and [here](https://sympa.inria.fr/sympa/arc/caml-list/2015-04/msg00044.html) is the 1st draft of a proposal for the organisation of META files with ppx extensions.

whitequark:
I can only say that all other ppx extensions I wrote (deriving, import, etc) add `ppx` field at toplevel and export the runtime as `.runtime`. (Which I'll make this PR conform to, in a moment).

darioteixeira:
That scheme was considered in the discussion, but it was quickly pointed out that it has several disadvantages:

- Libraries with optional ppx extensions, such as LWT, package them as a `ppx` sub-package. If deriving et al package the ppx at the top-level, we'll have an inconsistent ecosystem.
- How would you organise a package that offers both a Camlp4 and a ppx extension? It makes more sense to put the Camlp4 and ppx extensions in `syntax` and `ppx` sub-packages.  This has the advantage of being future-proof if an alternative macro mechanism comes along.
- By putting the ppx in a sub-package you are making the build process more transparent.

Obviously whatever scheme you adopt is your choice.  Still, it would be nice to know if you have any compelling arguments in favour of your proposal, particularly if they outweight the disadvantages above.

whitequark:
Packages that offer ppx extensions as a convenience could have a `.ppx` subpackage, packages whose main purpose is to provide a ppx extension could do it at toplevel. There isn't a burning need to have everything in an ideally consistent way, especially seeing as many people just won't care anyway (look at non-Alain, non-mine ppx packages on opam, for example.)

I don't really see anything more transparent here. Any META package can do whatever it wants, and you don't have a simple way to signal "this package follows a certain naming convention", so you'd have to check the documentation either way.

Drup:
I was waiting for the discussion to settle down. I do find one of bunzli's argument to be valid: 
> It seems to me that Dario's scheme, that is always have PKG.ppx, is more forward looking. We could perfectly imagine that a library in the future will implement its functionality using another mechanism (for example in a non-ugly way, using a macro system). 

I propose that, for now, we add the ppx field and a runtime one and makes the root package depends on both (in order to be backward compat) and indicate users to use sedlex.ppx. Is that ok ?

whitequark:
I guess you could argue that sedlex is not inherently dependent on ppx, so yeah.

darioteixeira:
@Drup's compromise solution seems reasonable.  @alainfrisch: what's your opinion?

alainfrisch:
I don't think this preserves backward compatibility, since the ppx subpackage is currently used for a different purpose. But I'm pretty sure this doesn't break anything, so I'm fine with it.  One needs to find a name for what is currently called ppx, or just drop it.

Drup:
Ah, yes, the internal library, forgot about that one.

I usually call it sedlex.ppx.internal.

PR #32, opened 2015-05-09
Making more use of ppx_metaquot in ppx_sedlex.ml.

cakeplus:
A simple refactoring that was done as a test drive for [ppx_metaquot](https://github.com/alainfrisch/ppx_tools).

Conclusion: the tool seems promising, though at the moment it lacks support for more kinds of quotations and antiquotations, such as for constant literals and identifiers. Also, there's no support for antiquote splicing (I couldn't extract the items of a tuple pattern without it). Perhaps I will contribute some improvements myself (if ppx proves to be a useful replacement for Camlp4 in the subsequent evaluations).

### DIFF
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -268,64 +268,68 @@ let err loc s =
 let regexp_of_pattern env =
   let rec char_pair_op func name p tuple = (* Construct something like Sub(a,b) *)
     match tuple with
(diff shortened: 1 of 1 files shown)


alainfrisch:
Sorry for the slooooow reply, and thanks for the contribution.  Nevertheless, I'm not overly convinced that the change makes the code much simpler to read.  Call me old-school, but I find it easier to read `Ppat_or (p1, p2)` than `[%pat? [%p? p1 ] | [%p? p2 ] ]`.

PR #36, opened 2015-12-05
Use $(MAKE) instead of hard coded "make"

slegrand45:
This patch fixes the installation of sedlex on FreeBSD with opam. It also seems that it's the recommended way for recursive make: https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable


alainfrisch:
Thanks!

PR #37, opened 2016-04-14
Fix for Sedlex failing to build on 4.03.

neolem:
This pull request fixes the issue of `sedlex` not building on `ocaml 4.03+beta2`. In addition, current sedlex version in opam is also failing due to the same issue 

A sample error we get is as follows,
```
File "ppx_sedlex.ml", line 79, characters 27-36:
Error: This variant expression is expected to have type Parsetree.constant
       The constructor Const_int does not belong to type Parsetree.constant
```
We get a few other errors all related to the `Parsetree.constant` variant being changed in `4.03`. 

`Parsetree.constant` variant has changed in `4.03`. See [Parsetree.constant 4.03](https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli#L20)

### DIFF
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -306,7 +306,7 @@ let regexp_of_pattern env =
         char_pair_op Sedlex.intersection "Intersect" p arg
     | Ppat_construct ({txt = Lident "Chars"}, arg) ->
         let const = (function 
-          | Some {ppat_desc=Ppat_constant (const)} -> Some (constant_type const)
+          | Some {ppat_desc=Ppat_constant (const)} -> Some (of_constant const)
           | _ -> None) arg 
         in 
         begin match const with
@@ -319,14 +319,14 @@ let regexp_of_pattern env =
         | _ -> err p.ppat_loc "the Chars operator requires a string argument"
         end
     | Ppat_interval (i_start, i_end) -> 
-        begin match constant_type i_start, constant_type i_end with
+        begin match of_constant i_start, of_constant i_end with
           | Pconst_char c1, Pconst_char c2 -> Sedlex.chars (Cset.interval (Char.code c1) (Char.code c2))
           | Pconst_integer(i1,_), Pconst_integer(i2,_) -> 
               Sedlex.chars (Cset.interval (codepoint (int_of_string i1)) (codepoint (int_of_string i2)))
           | _ -> err p.ppat_loc "this pattern is not a valid interval regexp"
         end 
     | Ppat_constant (const) -> 
-        begin match constant_type const with 
+        begin match of_constant const with 
           | Pconst_string (s, _) -> regexp_for_string s
           | Pconst_char c -> regexp_for_char c
           | Pconst_integer(i,_) -> Sedlex.chars (Cset.singleton (codepoint (int_of_string i)))
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -7,6 +7,7 @@ open Parsetree
 open Asttypes
 open Ast_helper
 open Ast_convenience
+open Ast_convenience.Constant
 
 module Cset = Sedlex_cset
 
@@ -309,7 +310,7 @@ let regexp_of_pattern env =
           | _ -> None) arg 
         in 
         begin match const with
-        | Some (Const_string(s,_))->
+        | Some (Pconst_string(s,_))->
             let c = ref Cset.empty in
             for i = 0 to String.length s - 1 do
               c := Cset.union !c (Cset.singleton (Char.code s.[i]))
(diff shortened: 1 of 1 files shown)


Drup:
Unfortunatly, we can't merge this like that, it wouldn't compile on 4.02. We should be able to avoid any preprocessing by using ppx_tool's Ast_convenience functions. Could you give it a try ?

alainfrisch:
What's the problem declaring that the previous released version of sedlex requires 4.02 and the next one 4.03?

Drup:
> What's the problem declaring that the previous released version of sedlex requires 4.02 and the next one 4.03?

It's not nice for users and we can avoid it without too much effort ?

rgrinberg:
Please don't dump 4.02.3 support so fast. The transition to 4.03 isn't going to be so smooth for all of us :(

neolem:
hmm ... @Drup ok, sure thing. I will give it a try later but not too hopeful though :). 

If we want to maintain older compilers then perhaps we can use something like `cppo`. This is used by `utop` to likely provide similar support to older compilers. [utop](https://github.com/diml/utop/blob/master/src/lib/uTop_main.cppo.ml#L868). `cppo` route maybe perhaps more easier, manageable. Thoughts???

Drup:
If you need conditional thing, propose them for inclusion in ppx_tools instead.

neolem:
@Drup sorry for the delay. `ppx_tools Ast_convenience` module and helper functions in that module are already updated to use `4.03` version `parsetree` and it seems it doesn't support `4.02` ocaml. So I am not sure if we can provide backwards compatibility without using something like `cppo`. Perhaps @alainfrisch @whitequark can add support for `4.02` parsetree in `ppx_tools`.

whitequark:
There's a 4.02 release of ppx_tools that supports it.

neolem:
Do you mean ppx_tools that supports both 4.02 and 4.03?

whitequark:
There are two different releases for 4.02 and 4.03, but they have compatible API.

neolem:
So, if we use ppx_tools 4.02, is it upwards compatible with 4.03? i.e. since opam works on source level, can we build both sedlex and ppx_tools(4.02) successfully with 4.03 compiler? I am not really familiar with ppx_tools yet ...

whitequark:
No, ppx_tools 4.02 is only compatible with OCaml 4.02, and ppx_tools 4.03 is only compatible with OCaml 4.03. However, they export the same interface (e.g. the Label module).

neolem:
@whitequark, yep, I thought so. I was looking at in particular [ast_convenience](https://github.com/alainfrisch/ppx_tools/blob/f908f407533609478fcdc15adef5c1aae8106e30/ast_convenience.ml). Is there an api/module that abstracts over the changes in `parsetree` from `4.02` to `4.03`? ,i.e. works with `parsetree` from ocaml `4.02` and `4.03` in the same package version.

We are looking at a mechnism to make sedlex build on both 4.02 and 4.03. Perhaps employing a  similar mechanism that [utop](https://github.com/diml/utop/blob/master/src/lib/uTop_main.cppo.ml#L868) implements to provide backwards compaitibilty for ocaml `<= 4.02` while at the same time supporting `4.03`? 

At the moment `sedlex` is refusing to compile on 4.03 due to api changes in `parsetree` from 4.02. what are our options to make sedlex compatible with `4.02` and `4.03` at the same time? Can ppx_tools provide package that supports both 4.02 and 4.03 at the same time? Is this in the works? Or should we go with cppo?

whitequark:
This is not the job of ppx_tools; use cppo.

neolem:
@Drup @alainfrisch It seems we need to use `cppo` if we want to provide backwards compatibility for 4.02 and be compatible with 4.03 in the same package. Thanks to @whitequark for confirming our choices. 
If  `cppo` is not an option then I think the approach suggested by Alain is the best of the options so far.
> ... declaring that the previous released version of sedlex requires 4.02 and the next one 4.03?

Drup:
> Is there an api/module that abstracts over the changes in parsetree from 4.02 to 4.03? ,i.e. works with parsetree from ocaml 4.02 and 4.03 in the same package version.

Yes, ppx_tools.

I feel like you are imposing meaningless weird constraints. Consider the last version of `lwt`, it has no cppo, it has constraint `ppx_tools > 1.0` and it compiles on both 4.02 and 4.03, using a different ppx_tools versions in each case.

neolem:
ah, okay. Just so that I am understanding you correctly, does this mean to  replace all uses of `parsetree` with ppx_tools `ast_convenience` helper functions? Interesting ... I will look into `lwt` more ... Thanks for the pointer.

neolem:
Hi @Drup, 

I have updated the pull request with the required changes based on our discussion. There are 2 other pull requests(both to ppx_tools(version 4.02 and 4.03) that I have had to send in order to support this pull request. 

`ppx_tools` pull requests
[4.03 #41 ](https://github.com/alainfrisch/ppx_tools/pull/41)
[4.02 #40](https://github.com/alainfrisch/ppx_tools/pull/40)

Thanks.

src/syntax/ppx_sedlex.ml
```diff
@@ -76,7 +76,7 @@ let decision_table p =
 
 
 let appfun s l = app (evar s) l
-let pint i = Pat.constant (Const_int i)
+(* let pint i = Pat.constant (Const_int i) *)
```
Drup: Just remove it. :)

Drup:
@neolem Nice. I'll merge when things are merged in ppx_tools are merged.

neolem:
@Drup ppx_tools has merged the required changes. This pull request can be merged now. Thanksl.

src/syntax/ppx_sedlex.ml
```diff
@@ -7,6 +7,7 @@ open Parsetree
 open Asttypes
 open Ast_helper
 open Ast_convenience
+open Ast_convenience.Constant
```
Drup: Don't use a global open for this please.

neolem:
@Drup Done. Removed the global open.

Drup:
Thanks!

whitequark:
@Drup so, are you going to merge this?

whitequark:
Actually, looks like it fails to build on 4.03 over here:
```
# File "ppx_sedlex.ml", line 308, characters 60-80:
# Error: Unbound module Constant
# Hint: Did you mean Const?
```

alainfrisch:
> Actually, looks like it fails to build on 4.03 over here:

You should use ppx_tools >= 5.0.

I've pushed the suggest fix (slightly adapted).

PR #38, opened 2016-05-03
upgrade to ocaml 4.03.0

zoggy:


alainfrisch:
Thanks!  However #37 seems better as it preserves compatibility with 4.02.

zoggy:
Ah, didn't see it. I let you forget this PR but you can still fix the META file.

alainfrisch:
> you can still fix the META file.

Indeed. ==> commit 27a44dc

Thanks!

PR #40, opened 2016-05-26
Fixes compilation for OCaml 4.03

modlfo:
These are the changes I made to compile in OCaml 4.03. However I don't know what's the best way of making the code compatible with older versions of OCaml.


modlfo:
I just saw there exists a few more pull request fixing this problem so I'm closing this one.

Issue #18, opened 2014-12-09
Add repeating patterns


whitequark:
For example, `Rep (pat, 3)`, meaning three repeats of `pat`, and `Rep (pat, 3, 5)`, meaning three to five. Corresponding to `{}` in regular expressions.

mroch:
this is done, right?

PR #42, opened 2016-12-03
add support for pattern r{n}/r{n, m}

jpathy:
implements #18 via `Rep (r, n)` for repeat n and `Rep (r, n .. m)` for repeat in range n to m.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,0 +1,5 @@
+PKG compiler-libs ppx_tools ppx_tools.metaquot
+
+SRC src/**
+
+B src/**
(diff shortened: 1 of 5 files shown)


alainfrisch:
Thanks for the contribution.  The code could be written more directly by calling Sedlex.seq directly instead of synthetizing some AST that will be analyzed directly.  Also, you can avoid the "accumulator"; there is no risk of blowing the stack here with a direct-style recursion.  As for checking the bounds, you should check, in addition to n1 <= n2, that 0 <= n1 (not that n2 <> 0).

jpathy:
Thanks for the suggestions. I have commited the changes.

alainfrisch:
Thanks!  I've pushed a slightly simplified implementation.

PR #43, opened 2017-03-17
Add lexing positions API.

toots:
This would make it much easier to use `sedlex` with `menhir` through their revised API. Please consider for inclusion!

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -153,12 +153,9 @@ let next lexbuf =
       else (refill lexbuf; lexbuf.buf.(lexbuf.pos))
     else lexbuf.buf.(lexbuf.pos)
   in
-  begin
-    match i with
-      (* '\n' = 10 *)
-      | 10 -> new_line lexbuf
-      | _ -> incr_curr lexbuf
-  end;
+  incr_curr lexbuf;
+  (* '\n' = 10 *)
+  if i = 10 then new_line lexbuf;
   if i = eof then lexbuf.finished <- true else lexbuf.pos <- lexbuf.pos + 1;
   i
 
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -52,7 +52,7 @@ type lexbuf = {
 
 let chunk_size = 512
 
-let empty_position () = {
+let empty_position = {
   Lexing.
     pos_fname = "";
     pos_lnum = 1;
@@ -60,7 +60,7 @@ let empty_position () = {
     pos_cnum = 0;
 }
 
-let empty_lexbuf () = {
+let empty_lexbuf = {
   refill = (fun _ _ _ -> assert false);
   buf = [| |];
   len = 0;
@@ -70,13 +70,13 @@ let empty_lexbuf () = {
   marked_pos = 0;
   marked_val = 0;
   finished = false;
-  lex_marked_p = empty_position ();
-  lex_start_p = empty_position ();
-  lex_curr_p = empty_position ()
+  lex_marked_p = empty_position;
+  lex_start_p = empty_position;
+  lex_curr_p = empty_position
 }
 
 let create f = {
-  empty_lexbuf () with
+  empty_lexbuf with
     refill = f;
     buf = Array.make chunk_size 0;
 }
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -43,11 +43,24 @@ type lexbuf = {
   mutable marked_val: int;
 
   mutable finished: bool;
+
+  (* Compat with other lexers. *)
+  mutable lex_marked_p: Lexing.position;
+  mutable lex_start_p:  Lexing.position;
+  mutable lex_curr_p:   Lexing.position
 }
 
 let chunk_size = 512
 
-let empty_lexbuf = {
+let empty_position () = {
+  Lexing.
+    pos_fname = "";
+    pos_lnum = 1;
+    pos_bol = 0;
+    pos_cnum = 0;
+}
+
+let empty_lexbuf () = {
   refill = (fun _ _ _ -> assert false);
   buf = [| |];
   len = 0;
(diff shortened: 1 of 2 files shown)


src/lib/sedlexing.ml
```diff
@@ -43,11 +43,24 @@ type lexbuf = {
...
 let chunk_size = 512
 
-let empty_lexbuf = {
+let empty_position () = {
+  Lexing.
+    pos_fname = "";
+    pos_lnum = 1;
+    pos_bol = 0;
+    pos_cnum = 0;
+}
```
Drup: `Lexing.dummy_pos` ?

toots: These need to be freshly instantiated but you can use other constructs if you prefer.

Drup: Why ? The content is not mutable.

toroidal-code:
I've done some work on something like this myself, setting up an ocamllex-like interface so that menhir could work with it cleanly. [Here's](https://github.com/toroidal-code/sedlex/commit/0f81344ddaa986b73a093f2149dd82895fa673be#diff-36e658c6bfa60cefbd5676a33946ca33R48) the commit if you'd like to take a look.

It's probably not up-to-date with the master, and I know I did some stuff that would break compatibility with the current version's API, so that would need to be fixed, but as I _know_ it works, it might at least help.

toots:
With these changes, plugging `sedlex` into `menhir` becomes fairly easy:

```
let menhir_tokenizer lexbuf = fun () ->
  let token = tokenizer lexbuf in
  let (start_p,end_p) = Sedlexing.positions lexbuf in
  (token lexbuf,start_p,end_p)
in
let parser =
  MenhirLib.Convert.Simplified.traditional2revised parser
in
parser tokenizer
```

rgrinberg:
Having some official documentation on how to integrate with menhir would be very nice. I remember running into this problem as well.

src/lib/sedlexing.ml
```diff
@@ -117,26 +133,46 @@ let refill lexbuf =
...
 let next lexbuf =
   let i =
     if lexbuf.pos = lexbuf.len then
       if lexbuf.finished then eof
       else (refill lexbuf; lexbuf.buf.(lexbuf.pos))
     else lexbuf.buf.(lexbuf.pos)
   in
+  incr_curr lexbuf;
+  (* '\n' = 10 *)
+  if i = 10 then new_line lexbuf;
```
clouds56: Handle `new_line` automatically on `next` is great, but why only '\n' = 10 would be treated as newline and not configurable? see also [wiki](https://en.wikipedia.org/wiki/Newline#Unicode).
Is it a option to have a mutable field for list of possible line_end characters?

clouds56:
I wonder in this case, what's the difference between `lexbuf.pos + lexbuf.offset` and `lexbuf.lex_curr_p.pos_cnum`, and if they are the same, should we remove one of them? (maybe only `lex_curr_p` and `offset` would be OK)

src/lib/sedlexing.ml
pmetzger: Generally speaking I'd say "\r\n" and "\n" (and only those!) should be treated as newlines by default. The various Unicode flavors out there are not particularly recommended for real world use. Some of them exist just to map chars present in historical character sets. However, I do agree that allowing the user to configure this would be reasonable. Perhaps after this gets integrated we could improve it and request a pull?

alainfrisch: I believe the extra work required at each character (allocating a record + a few memory reads) could slow down the lexers significantly.  See https://github.com/ocaml/ocaml/pull/1590 and https://github.com/ocaml/ocaml/pull/1585.

Also, is the code correct (incr_curr vs lex_p_incr_curr)?

clouds56: No, I think it's a topo, and should be `lex_p_incr_curr`.

Do you mean we should not update lex_curr_p in `next` but generate it when calling `lexing_positions`, `mark` and `backtrace`? (just like what #58 do)

What about new_line detection? I think it should be in `next` to get most accurate column info.

alainfrisch: I'm just saying that automatically updating the field has a cost, and some application are not interesting in tracking line numbers and might not want to pay the extra cost.

One solution is to do like in #1590, i.e. update positions only if they are not `== dummy_pos`.

clouds56: @pmetzger Yes we are doing so.

I also don't know how much the cost is. If you meet some performance issue or would be willing to do some performance test after checked in, we could try to do some work around it.

clouds56:
I could not modify this PR, so I created a new one #59

src/lib/sedlexing.ml
alainfrisch: To give an idea of the cost, for ocamllex:

https://github.com/ocaml/ocaml/pull/1585/commits/bbbfefc1214dd490650538b9d5558e0b67545342

In that benchmark, lexing throughput moved from 87 Mb/s to 134 Mb/s.

pmetzger: > In that benchmark, lexing throughput moved from 87 Mb/s to 134 Mb/s.

That seems like a shame. When I'm writing lexers in C I never have overhead like that for something small like tracking line numbers. Is it possible that the compiler might be able to optimize this case better with careful coding, say by doing something intelligent about the record allocation?

clouds56: Not possible with current workflow I think. (correct me if I'm wrong)
When we change the `lex_curr_p`, it is possible referenced also by `lex_mark_p`, so compiler would have no idea in compile time if it could reuse the record and have to create a new one and wait the old one to GC.

My concern is if the cost is mainly on memory allocation and GC. If so, could we make `lex_curr_p` and `lex_mark_p` plain in lexbuf to avoid them?

alainfrisch: > If so, could we make lex_curr_p and lex_mark_p plain in lexbuf to avoid them?

I believe this would probably help a lot.

Issue #48, opened 2017-05-29
Move the build system to jbuilder?


bikallem:
@alainfrisch @Drup Is there any interest in moving the current build system of `sedlex, ppx_tools, gen` to jbuilder? I will do the migration.

alainfrisch:
ppx_tools: I'd rather not switch unless there is a strong argument in favor. I prefer to minimize dependencies of this package as much as possible.

sedlex: I don't really care; I'll let other maintainers comment.

gen: not under my control.

rgrinberg:
Note that people should never use ppx_tools anymore unless they really know what they're doing. Ppx_tools_versioned should almost always be preferred as it will usually cause less grief for users.

That being said, I still think that ppx_tools should switch to jbuilder when installed from opam. jbuilder is much better at building and installing all the correct files and their variations (bytecode vs native, cmti, mli's, etc.) But that doesn't mean that ppx_tools can't keep its make based build system for those who would like to minimize dependencies.

@alainfrisch What do you think? Can we add a jbuild file just for opam to use?

alainfrisch:
For ppx_tools, ok for add a jbuild file, but be prepared to be asked to update it!  (Although I hope people will switch to ppx_tools_versioned.)

For sedlex, there is also #46

Issue #46, opened 2017-03-30
Using sedlex with jbuilder?


adrieng:
I'd like to use sedlex with jbuilder, however jbuilder does not support normal ppx rewriters for performance reasons: https://github.com/janestreet/jbuilder/issues/14

As mentioned in the discussion above, one solution is to port sedlex to use the [ocaml-migrate-parsetree](http://github.com/let-def/ocaml-migrate-parsetree) driver API. One other would be to use [ppx_driver](http://github.com/janestreet/ppx_driver), but that looks like more work (and dependencies?) to me.

I actually did the porting work, so if anybody is interested, I will submit a pull request.

alainfrisch:
Yes, thanks, a PR to switch to jbuilder will be appreciated!

smolkaj:
Has this been fixed? I'm using the following ugly workaround for now, but it's probably brittle:
```
(rule
 ((targets (Lexer.ml))
  (deps    (Lexer.sedlex.ml))
  (action  (run ${ocaml_where}/../sedlex/ppx_sedlex ${<} -o ${@}))))
```

PR #54, opened 2017-11-28
port to jbuilder

smolkaj:
This addresses  #48 and #46.


META
```diff
@@ -1,14 +0,0 @@
...
-requires = "gen"
-ppx = "./ppx_sedlex --as-ppx"
-
-package "ppx" (
-  version = "1.99.4"
-  description = "ppx implementation of sedlex"
-  requires = "ocaml-migrate-parsetree ppx_tools_versioned gen"
-  archive(byte) = "sedlex.cma"
-  archive(native) = "sedlex.cmxa"
-)
```
smolkaj: META file is no longer needed

.gitignore
```diff
@@ -8,6 +8,9 @@
 *.cmxs
 *.cmxa
 *.exe
+*.install
+*.merlin
+_build
 libdoc
 src/syntax/ppx_sedlex
 examples/complement
```
smolkaj: Don't commit auto generated files

src/lib/Makefile
```diff
@@ -1,26 +0,0 @@
...
-
-doc:
-	rm -rf ../../libdoc
-	mkdir ../../libdoc
-	ocamlfind ocamldoc -package gen -html sedlexing.mli -d ../../libdoc
-
-
-
-clean:
-	rm -f *~ *.cm* *.a *.lib *.exe *.o *.obj
```
smolkaj: Makefile no longer needed

src/syntax/Makefile
```diff
@@ -1,54 +0,0 @@
...
-ppx_sedlex.opt$(EXE): sedlex.cmxa
-	$(OCAMLOPT) \
-		-predicates ppx_driver \
-		-o $@ -linkpkg -linkall \
-		-package ppx_tools_versioned \
-		$< \
-		$(MIGRATE_PARSETREE_MAIN_NATIVE)
-
-clean:
-	rm -f *~ *.cm* *.a *.lib *.o *.obj *.annot ppx_sedlex$(EXE) ppx_sedlex.opt$(EXE)
```
smolkaj: Makefile no longer needed

sedlex.opam
```diff
@@ -5,9 +5,13 @@ maintainer: "Alain Frisch <alain.frisch@lexifi.com>"
...
 ]
-build: [ [make "all"]  [make "opt"] ]
-install: [make "install"]
-remove: [["ocamlfind" "remove" "sedlex"]]
+
+build: ["jbuilder" "build" "-p" name "-j" jobs]
+build-doc: ["jbuilder" "build" "@doc" "-p" name "-j" jobs]
+build-test: ["jbuilder" "runtest" "-p" name "-j" jobs]
+install: ["jbuilder" "install" "-p" name "-j" jobs]
+remove: ["jbuilder" "uninstall" "-p" name "-j" jobs ]
```
rgrinberg: There's no need for `install:` or `remove:` here. This is because jbuilder will generate a .install file which is enough for opam to know how to do installation/uninstallation.

smolkaj: Thanks. Fixed.

smolkaj:
@alainfrisch: Any chance this could be merged into master in the near future? Here are two reasons why that would be great:
* This makes sedlex compatible with jbuilder-based projects, which is starting to become the defacto default 
* My pull request actually reduces complexity

Drup:
I haven't tested, but it looks alright. I think you can be even more agressive with the makefile cleanup. The distrib stuff is probably not needed anymore.

smolkaj:
@Drup: Sounds good 👍 Done.

alainfrisch:
> @alainfrisch: Any chance this could be merged into master in the near future?

I've no opposition, but since I know very little about jbuilder, I'll let @Drup make the call.

Makefile
```diff
@@ -2,73 +2,26 @@
...
-	rm -rf libdoc
-
-test: clean all opt
-	cd examples && $(MAKE) test
-
-INSTALL=META src/syntax/sedlex.cma src/syntax/ppx_sedlex$(EXE) src/lib/sedlexing.cma src/lib/sedlexing.cmi
-
-INSTALL_OPT=src/syntax/sedlex.cmxs src/syntax/sedlex$(EXT_LIB) src/syntax/sedlex.cmxa src/syntax/ppx_sedlex.opt$(EXE) src/lib/sedlexing.cmx src/lib/sedlexing$(EXT_LIB) src/lib/sedlexing.cmxa
+build:
+	time -p jbuilder build @install
```
Drup: Please remove the call to time, it's not useful.

smolkaj: Done.

Drup:
Thanks!

rgrinberg:
The opam file doesn't include jbuilder as a dependency FYI. It should include it as a build dependency instead of findlib.

smolkaj:
Good point! I opened another PR to fix this.

PR #55, opened 2017-12-08
Addendum to PR #54: build now depends on jbuilder

smolkaj:
As @rgrinberg pointed out in #54, the opam file should specify the build dependency on jbuilder.


smolkaj:
Also, I would be grateful if one of the maintainers could push a fresh release to opam, now that sedlex is compatible with the jbuilder ecosystem.

rgrinberg:
Yes, please. An announcement on discuss.ocaml.org wouldn't hurt either :)

PR #58, opened 2018-04-03
Add support/helping module for menhir

clouds56:
inspired from [sedlex_menhir.ml](https://github.com/unhammer/ocaml_cg_streamparse/blob/master/sedlex_menhir.ml) and original [here](https://github.com/Drup/llvm/blob/3c43000f4e86af5b9b368f50721604957d403750/test/Bindings/OCaml/kaleidoscope/src/syntax.ml)


clouds56:
Minimal example (add sedlex.menhir to dependencies)

```ocaml
open Sedlexing_menhir
open Parser
let file = "some.txt"
let () =
  let c = open_in file in
  let lb = Sedlexing.Utf8.from_channel c in
  try
    let ast = sedlex_with_menhir ~file Parser.prog Lexer.token lb in
    close_in c;
    dump ast
  with
    | Sedlexing_menhir.ParseError e ->
        report (current_position lb);
        eprintf "Parse error@.";
        exit 1
```

compared to original

```ocaml
let lb = Lexing.from_channel c in
let ast = Parser.prog Lexer.next_token lb
```

we could use sedlex more flexible with menhir

sedlex.opam
```diff
@@ -15,6 +15,7 @@ depends: ["jbuilder" {build}
           "ocaml-migrate-parsetree"
           "gen"
          ]
+depopts: ["menhir"]
```
alainfrisch: I know little about opam, but I was under the impression that optional dependencies were not trendy these days.  (In particular, if menhir is installed after sedlex, one would need to recompile the whole sedlex packing only the get the extra module.)  Wouldn't it be better to ship the new module as a separate opam package?

alainfrisch:
How does this compare to #43?

alainfrisch:
Assigning to @Drup, who started to comment on #43.

clouds56:
@alainfrisch, I've not seem #43 before I created the PR.

I think the basic idea is the same, this patch makes minimal modification to the core `lexbuf`, and provided a **optional** helping module that connected to menhir (it would be build only menhir is installed).

I think it necessary to add position support in sedlex, either this or #43 should be fine.

sedlex.opam
clouds56: Fixed.
Though I like the idea that opam would auto rebuilt this package for new dependency menhir. I think not everybody would like to search and install another package for just a few lines of code (but it is fair that rebuilt may cost especially if some large package would depend on this)

Drup:
I think I prefer the API in #43: 
- It works with `Lexing.position` right from the start, instead of going through `(int * int * int)`.
- It doesn't hardcode the link with menhir, it just make the API more ocamllex-like, which I feel is better in the long run.

All the function you add in `Sedlexing_menhir` are nice, but there is no reason for them to be there, you could directly put them in the main module, and drop the dependency to menhir, since it's basically a 4 lines boilerplate that can be documented.

clouds56:
That's fine, so what's blocking #43 from checking in? If there's some issue I'm happy to help.

I thought you dislike making dependency to Lexing so I try to return `(int * int * int)` and later convert into `Lexing.position` in another module.
The module Sedlexing_menhir would be a separate package so no additional dependency of menhir would be included for sedlex. It is up to users to use sedlex_menhir or write it themselves.

If choose not to have `Sedlexing_menhir`, I think a comprise could be add an API like `?file:string -> lexbuf -> 'token * Lexing.position * Lexing.position`

Drup:
> That's fine, so what's blocking #43 from checking in? If there's some issue I'm happy to help.

Honestly? I simply forgot about it ... :/
Since you seem motivated to work on this, we can take the occasion to push it a bit more than what @toots  did. Also, I have a bit of time to review things now!

> I thought you dislike making dependency to Lexing

`Lexing` is in the stdlib, so no, that's not an issue at all.

> If choose not to have Sedlexing_menhir, I think a comprise could be add an API like ?file:string -> lexbuf -> 'token * Lexing.position * Lexing.position

Yes, I like that option a lot more. Let's provide 1) function that returns `Lexing.position` 2) Functions that users can pass easily to `traditional2revised`, and have a nice documentation that show off the two lines of code needed to plug in menhir.

clouds56:
Close in favor of #59

PR #59, opened 2018-04-05
Add lexing compatible position info

clouds56:
fix #43 and #58

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -212,7 +212,7 @@ let lexing_positions lexbuf =
   } in
   (start_p, curr_p)
 
-let convert_for_menhir lexer' lexbuf =
+let with_tokenizer lexer' lexbuf =
   let lexer () =
     let token = lexer' lexbuf in
     let (start_p, curr_p) = lexing_positions lexbuf in
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -153,12 +153,9 @@ let next lexbuf =
       else (refill lexbuf; lexbuf.buf.(lexbuf.pos))
     else lexbuf.buf.(lexbuf.pos)
   in
-  begin
-    match i with
-      (* '\n' = 10 *)
-      | 10 -> new_line lexbuf
-      | _ -> incr_curr lexbuf
-  end;
+  incr_curr lexbuf;
+  (* '\n' = 10 *)
+  if i = 10 then new_line lexbuf;
   if i = eof then lexbuf.finished <- true else lexbuf.pos <- lexbuf.pos + 1;
   i
 
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -183,6 +183,7 @@ let backtrack lexbuf =
   lexbuf.marked_val
 
 let rollback lexbuf =
+  lexbuf.lex_curr_p <- lexbuf.lex_start_p;
   lexbuf.pos <- lexbuf.start
 
 let lexeme_start lexbuf = lexbuf.start + lexbuf.offset
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -52,7 +52,7 @@ type lexbuf = {
 
 let chunk_size = 512
 
-let empty_position () = {
+let empty_position = {
   Lexing.
     pos_fname = "";
     pos_lnum = 1;
@@ -60,7 +60,7 @@ let empty_position () = {
     pos_cnum = 0;
 }
 
-let empty_lexbuf () = {
+let empty_lexbuf = {
   refill = (fun _ _ _ -> assert false);
   buf = [| |];
   len = 0;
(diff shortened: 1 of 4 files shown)


clouds56:
If https://github.com/ocaml/ocaml/pull/794 checked in, we could make advantage of new feature `[@unboxed]`.
For now just duplicate these `bol` and `line` fields.

src/lib/sedlexing.ml
```diff
@@ -117,48 +140,84 @@ let refill lexbuf =
...
+    pos_bol = lexbuf.curr_bol;
+  } in
+  (start_p, curr_p)
+
+let convert_for_menhir lexer' lexbuf =
+  let lexer () =
+    let token = lexer' lexbuf in
+    let (start_p, curr_p) = lexing_positions lexbuf in
+    (token, start_p, curr_p)
+  in lexer
```
Drup: What happens at the end of stream ? Does the menhir API specifies something ?

src/lib/sedlexing.mli
```diff
@@ -136,6 +148,9 @@ val backtrack: lexbuf -> int
     internal slot of the buffer, and performs backtracking
     (the current position is set to the value of the backtrack position). *)
 
+val convert_for_menhir: (lexbuf -> 'token) -> lexbuf -> (unit -> 'token * Lexing.position * Lexing.position)
+(** [backtrack convert_for_menhir] convert a lexer to a rivised lexer that
+    menhirLib accepts *)
```
Drup: I think a better (and more neutral) name would with `with_tokenizer`. The documentation could look like

```
[with_tokenizer tokenizer lexbuf] returns a generator of tokens annotated with positions. 
This generator can be used with {!TheRightMenhirFunction}
```

src/lib/sedlexing.mli
```diff
@@ -48,6 +48,14 @@ val create: (int array -> int -> int -> int) -> lexbuf
         position [pos], and return the number of characters provided. A
         return value of 0 means end of input. *)
 
+val set_curr_p: lexbuf -> Lexing.position -> unit
+    (** set initial [lex_curr_p] for [lexbuf],
+        if set to [Lexing.dummy_pos], Sedlexing would not track position info for you *)
+
+val set_filename: lexbuf -> string -> unit
+    (** set [lex_curr_p.pos_fname] for [lexbuf],
+        this implies to track the position info *)
```
Drup: Try to start documentation comments with the function. For example here it would be: `[set_filename lexbuf file] sets <bla> to <bla>.`

src/lib/sedlexing.ml
```diff
@@ -117,48 +140,84 @@ let refill lexbuf =
   end
   else lexbuf.len <- lexbuf.len + n
 
+let new_line lexbuf =
+  if lexbuf.curr_line != 0 then
+  lexbuf.curr_line <- lexbuf.curr_line + 1;
+  lexbuf.curr_bol <- lexbuf.pos + lexbuf.offset
+
```
Drup: This function should probably be exposed in the API.

clouds56: Should it? If so, someone from ocmallex might be confused if they expect `next` to **not** handle new_line automatically and use this incorrectly.
Maybe we should expose it later with some other api that provide ability to config line_ending. Set line_ending to empty list indicates handle the line ending by user. (And it worth another PR)

Drup: Fair enough. In all cases, the fact that line handling is automatic should be documented!

src/lib/sedlexing.ml
clouds56: In our package, eof has length 0 at file end, which is consist of ocamllex.
menhir API says only that position information is compatible with ocamlyacc and I could not found document of ocamlyacc about this.

clouds56:
I tested running 100 times of `String.make 1000000 ' '`

string | inline | record | no postion
------|-----|-----|-----
*base*|1.035044s|1.079814s|1.070310s
`ch = ' '`|1.863023s|2.127180s|1.632510s
`ch = '\n'`|2.005856s|2.302125s|1.606208s
dummy_pos|2.033207s|2.253284s|n/a

code of *base*
```ocaml
  let ch = ' '
  let long_str = String.make 1000000 ch in
  let token_from str =
    let lexbuf = Sedlexing.Latin1.from_string long_str in
    (* let () = Sedlexing.set_curr_p lexbuf Lexing.dummy_pos in *)
    (* token lexbuf *) ()
  in time token_from long_str
```
where the dummy_pos means `ch = '\n'` and uncomment the line `Sedlexing.set_curr_p`

clouds56:
@Drup, any issues now?

src/lib/sedlexing.mli
```diff
@@ -86,6 +94,14 @@ val lexeme_length: lexbuf -> int
...
         string. *)
 
+val lexing_positions : lexbuf -> Lexing.position*Lexing.position
+    (** [Sedlexing.lexing_positions lexbuf] returns the start and current
+        positions with type [Lexing.position], suitable for consumption
+        by parsers like [Menhir]. *)
+
+val new_line: lexbuf -> unit
+    (** [Sedlexing.new_line lexbuf] tolds lexbuf that  *)
+
```
Drup: that ? :)

Drup:
I think the code is fine. I'll merge it, and fix documentation later on.

src/lib/sedlexing.mli
clouds56: Sorry, I must forget finishing the document at that time (it's tiresome writing document, isn't it :))

Issue #45, opened 2017-03-22
Interactive mode


toots:
Hi,

I'm running into an issue migrating an existing app using `ocamllex`/`ocamlyacc` to `sedled`/`menhir`. The problem at hand here does concern only `sedlex` as I've successfully tested it with `ocamllex`/`menhir`.

The application has an interactive mode, it works this way:
`parser.mly`:
```
%token <string> NUMBER
%token SEQSEQ EOF

%start interactive
%type <int> interactive

%%

exprs:
  | NUMBER { int_of_string $1 }

interactive:
  | exprs SEQSEQ { $1 }
  | EOF { raise End_of_file }
```

`lexer.ml`:
```
open Parser

let rec tokenizer lexbuf =
  match%sedlex lexbuf with
    | '\n' -> tokenizer lexbuf
    | Plus('0'..'9') -> NUMBER (Sedlexing.Utf8.lexeme lexbuf)
    | ';', ';' -> SEQSEQ
    | eof -> EOF
    | _ -> failwith "Parse error!"
```

`main.ml`:
```
let interactive =
  MenhirLib.Convert.Simplified.traditional2revised Parser.interactive

let get_token lexbuf =
  let tokenizer () =
    (Lexer.tokenizer lexbuf, Lexing.dummy_pos, Lexing.dummy_pos)
  in
  interactive tokenizer

let () =
  let lexbuf = Lexing.from_channel stdin in
  let rec f () =
    Printf.printf "Got number from stdin: %d\n%!" (get_token lexbuf);
    f ()
  in
  f ()
```

When compiled using `ocamllex` and run through a terminal console, I can do:

```
% ./test
<type: 123;;\n>
Got number from stdin: 123
<type: 456;;\n>
Got number from stdin: 456
```

However, nothing happen when I do the same with the `sedlex` lexer. But it works when I do this:
```
% echo "123;;\n456;;\n" | ./test_sedlex
Got number from stdin: 123
Got number from stdin: 456
Fatal error: exception End_of_file
```

I'd imagine that this is most likely due to the difference in treating `ascii` characters vs. `utf8` characters. However, do y'all think that there is a way to mitigate this?

Thanks!

xguerin:
Seconded. I was expecting using a generator to address that problem, to no avail; `match%sedlex` only starts lexing the input when EOF is found. That also makes parsing from a pipe for instance quite problematic.

pmetzger:
[...]

PR #68, opened 2018-10-04
Embed chunk size in the lexbuf to allow interactive capture #45

xguerin:
Make the size of the refill chunk configurable. Enable interactive mode by using a size of 1 instead of the default 512.

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -74,10 +74,10 @@ let empty_lexbuf = {
 
 let create ?(chunk_size=512) f = {
   empty_lexbuf with
-    refill = f;
-    chunk_size;
-    buf = Array.make chunk_size (Uchar.of_int 0);
-    curr_line = 1;
+  refill = f;
+  chunk_size;
+  buf = Array.make chunk_size (Uchar.of_int 0);
+  curr_line = 1;
 }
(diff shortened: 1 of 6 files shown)


alainfrisch:
Some random comments:

  - You should share the default chunk size to avoid copying the literal 512 in many places.
  - Why is chunk_size declared mutable (there does not seem a way to change it after creation)?
  - You should document the meaning of the new chunk_size parameters in the interface.
  - Any idea on the impact on performances?

pmetzger:
Lexers can be made interactive without making the chunk size 1, so I'm not sure this is the right way to achieve the stated goal.

xguerin:
@pmetzger AFAICT the path to the stated goal is to either (1) remove buffering or (2) implement line-based buffering. Both solutions require re-architecting the input model currently used. Altering the `chunk_size` is the most direct way to implement (1) in the current state of the code. Performance was not a part of the deciding arguments as the modification is mostly intended for actual user-driven interactive use.

That being said I must agree that getting rid of the buffer is sub-optimal. I originally intended to make the buffer management parametric but the task seemed a lot more time-consuming than what I was ready to commit at the time. I don't mind going back to the drawing board and coming up with a more elegant solution.

PR #70, opened 2018-10-27
Restore OCaml 4.02.3 support

ELLIOTTCABLE:
For now, this just adds 4.02.3 back into the support-matrix, then I hope to try and resolve any problems that Travis surfaces.

Given that the local `.opam` was never updated to indicate differently, if this works out, there should be no further local changes necessary — we can just ask opam-repository to revert ocaml/opam-repository#12589.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
   global:
   - PACKAGE=sedlex
   matrix:
+  - OCAML_VERSION=4.02.3
   - OCAML_VERSION=4.03
   - OCAML_VERSION=4.04
   - OCAML_VERSION=4.05
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
   global:
   - PACKAGE=sedlex
   matrix:
-  - OCAML_SWITCH=4.02.3
+  - OCAML_VERSION=4.02 OCAML_SWITCH=4.02.3
   - OCAML_VERSION=4.03
   - OCAML_VERSION=4.04
   - OCAML_VERSION=4.05
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
   global:
   - PACKAGE=sedlex
   matrix:
-  - OCAML_VERSION=4.02.3
+  - OCAML_SWITCH=4.02.3
   - OCAML_VERSION=4.03
   - OCAML_VERSION=4.04
   - OCAML_VERSION=4.05
--- a/sedlex.opam
+++ b/sedlex.opam
@@ -14,6 +14,7 @@ depends: ["jbuilder" {build}
           "ppx_tools_versioned"
           "ocaml-migrate-parsetree"
           "gen"
+          "uchar"
          ]
 available: [ ocaml-version >= "4.02.3" ]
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -6,11 +6,12 @@ env:
...
+  - OCAML_VERSION=4.02 OCAML_SWITCH=4.02.3
   - OCAML_VERSION=4.03
-  - OCAML_VERSION=4.04
-  - OCAML_VERSION=4.05
-  - OCAML_VERSION=4.06
-  - OCAML_VERSION=4.07
+  # - OCAML_VERSION=4.04
+  # - OCAML_VERSION=4.05
+  # - OCAML_VERSION=4.06
+  # - OCAML_VERSION=4.07
```
pmetzger: Why did you comment out 4.04 to 4.07?

sedlex.opam
```diff
@@ -14,6 +14,7 @@ depends: ["jbuilder" {build}
           "ppx_tools_versioned"
           "ocaml-migrate-parsetree"
           "gen"
+          "uchar"
```
pmetzger: That's only a dependency on the older versions of course.

ELLIOTTCABLE: Is there a way to express that on opam? Also, isn't the compatibility library a no-op on newer versions?

Happy to fix, just not sure how to usefully change it.

.travis.yml
ELLIOTTCABLE: It was temporary, because I expected to need to use Travis to experiment with 4.02/4.03 support differences, and the long development loop (an hour or so) that that imposed was getting in my way. Bit of a hack; but I reverted it when I was done — hence WIP in the pull-request title!

ELLIOTTCABLE:
Should be good to go now! (Unless there's a way to improve / restrict the opam dependency on the `Uchar` compatibility-library.)

ELLIOTTCABLE:
I'm using my BuckleScript build, which is implicitly 4.02.3; but I haven't used it with the stock OCaml compiler. I can build the examples, at least, tonight!

ELLIOTTCABLE:
Also, I tried to add version-constraints to the dependency in e11b5, with no luck. I got the error reproduced below. I don't know how, precisely, to do that in an opam-1.2-compatible way. Should we upgrade the `.opam` file to opam 2.0? ¯\_(ツ)_/¯

```
[WARNING] Failed checks on sedlex package definition from source at file:///tmp/opam-ci-192:
             error 29: Package dependencies mention package variables: "ocaml:version"
Errors.
    error 29: Package dependencies mention package variables: "ocaml:version"
```

Drup:
Uchar is the empty package for recent ocaml versions, there is no reason whatsoever to play dependency games. :)

pmetzger:
Okay, if @Drup is comfortable with this, I am. @Drup, should we merge it?

Drup:
@pmetzger when CI is back in order yes. All the lines are still commented.

ELLIOTTCABLE:
@Drup sounds good! I was dicking around, trying to play said dependency-games. Glad I can dispense with that.

Restoring CI now; you can merge, assuming my next commit passes!

PR #73, opened 2019-03-17
Avoid warning about unused recursive flag in the generated code.

havleoto:
Sedlex always marks the function generated from [%sedlex match ...] with the recursive flag. In some cases, the function is actually not recursive and the generated code triggers a the compiler warning 39 "unused rec flag".

This patch computes the correct recursive flag and enables the warning for the example programs.


Drup:
Nice! Looks good.

PR #77, opened 2019-04-30
Buffer channel input, return None at the end of partial reads.

toots:
Finally came up with the proper fix for #45 !

This PR adds proper buffering on channel reads while also inserting a `None` entry at the end of each partial read. This allow the `fill_buf_from_gen` function to return without reaching blocking read on the buffer.

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -5,11 +5,32 @@
 exception InvalidCodepoint of int
 exception MalFormed
 
+let chunk_size = 512
 
+(* This function buffers channel read and inserts one [None] entry at the 
+ * end of each partial read, to allow the [refill] function to return, 
+ * preventing subsequent blocking read on the channel. *)
 let gen_of_channel chan =
-  let f () =
-    try Some (input_char chan)
-    with End_of_file -> None
+  let buf = Bytes.create chunk_size in
+  let cached = ref (-1) in
+  let position = ref (-1) in
+  let rec f () =
+    match !position, !cached  with
+      | _, 0 ->
+         None
+      | -1, _ ->
+         position := 0;
+         cached := input chan buf 0 chunk_size;
+         f ()
+      | len, c when len = c ->
+         position := -1;
+         if len = chunk_size then
+           f ()
+         else
+           None
+      | len, _ ->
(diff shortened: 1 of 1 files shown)


pmetzger:
It is probably only necessary to be careful about blocking on interactive uses; for non-interactive uses it probably slows things down a bit (and that can be significant when compiling a large code base; often lexing is one of the most time-consuming steps). I wonder if we should have two separate functions for this. Several other lexical analyzer generators have specific options for interactive vs. non-interactive use. Perhaps @Drup has an opinion.

toots:
A separate function could work too. I'm already doing that in the app itself.

Drup:
@pmetzger I'm not so convinced that having two modes is important. We would need benchmarks to really make a decision.

pmetzger:
@Drup A benchmark is probably a good idea.

pmetzger:
@toots I noticed this one was still open. What should we do about it?

toots:
Thanks for bringing this one up!

We have used this function in our application for a while and it's working pretty well. I say we could either merge it as it is or, if we can to be conservative with features, add those changes as a `from_interactive_channel`.

pmetzger:
Do you think you could do a quick benchmark to see if we should bother with a distinct mode?

hhugo:
Running a quick benchmark, it seems that this PR is actually faster than the implementation on master, probably due to the fact that we perform less c calls.

That said, I cannot convince myself that the implementation is correct. In particular, I don't understand how we can be sure that we don't return `None` at the very start of the `refill` call (which would look like an EOF)

hhugo:
Also, I think we would have invalid results if a give-back-control `None` was returned in the middle of Utf8.Helper.from_gen.

hhugo:
I have an alternative fix for #45, see #124

hhugo:
This can be closed @toots

toots:
Indeed! @hhugo do you have other pending changes? I'd say we should do a release soon now.

hhugo:
I don't have other changes planned. I wanted to run some more tests with jsoo first maybe.

hhugo:
@toots, I run tests with jsoo, all good. I've just open one last fix #126

PR #79, opened 2019-05-30
Fix typo in documentation for `lexeme_length`

MattWindsor91:
The ocamldoc attached to `Sedlexing.lexeme_length` starts with a reference to `Sedlexing.loc`, the function above it.  This looks like a straight copy-and-paste typo; this proposed commit directly replaces it with lexeme_length (but doesn't do any further adjusting).


pmetzger:
This was too obviously correct not to merge. Thanks for the fix, @MattWindsor91!

PR #110, opened 2022-03-28
Ppxlib.0.26.0 compatibility

pitag-ha:
This is a patch PR to make the PPX compatible with ppxlib.0.26.0 which has bumped the AST to 4.14/5.00.

### DIFF
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -284,15 +284,13 @@ let regexp_of_pattern env =
         List.fold_left (fun r p -> Sedlex.seq r (aux p))
           (aux p)
           pl
-    | Ppat_construct (_, Some (_ :: _, _)) ->
-        err p.ppat_loc "Explicit binders for type variables not supported"
-    | Ppat_construct ({txt = Lident "Star"}, Some ([], p)) ->
+    | Ppat_construct ({txt = Lident "Star"}, Some (_, p)) ->
         Sedlex.rep (aux p)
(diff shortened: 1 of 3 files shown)


src/syntax/ppx_sedlex.ml
```diff
@@ -284,13 +284,15 @@ let regexp_of_pattern env =
         List.fold_left (fun r p -> Sedlex.seq r (aux p))
           (aux p)
           pl
-    | Ppat_construct ({txt = Lident "Star"}, Some p) ->
+    | Ppat_construct (_, Some (_ :: _, _)) ->
+        err p.ppat_loc "Explicit binders for type variables not supported"
```
kit-ty-kate, line 288: Maybe same comment as in https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/142#issuecomment-1080543029

pitag-ha: Hey @kit-ty-kate, thanks for the review! See my answer [here](https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/142#issuecomment-1080635196). If you're sure that the good new behavior here is to ignore the new feature, let me know and I update the PR! (I'm asking because of your "maybe")

toots:
Thanks for the fix! I didn't have much time to think about this but it seems to me that most of the pattern matching involving this ppx shouldn't be making use of universal type bindings like the ones introduced in the AST changes so this seems safe to ignore them for now.

Would be curious to see if we could build up an example saying otherwise..

pitag-ha:
Thanks for adapting the PR, @toots!

Issue #102, opened 2021-08-12
Don't have `as` anymore 


chengtie:
I'm new to sedlex. I'm translating a lexer in ocamllex to sedlex. By ocamllex, I have e.g.,

```
rule token = parse
  | FLOAT as float "%"           { DOUBLE ((float_of_string float) /. 100. ) } 
```

Now, I need to use some string operations to achieve it:

```
let rec token buf =
  match%sedlex buf with
    | FLOAT, "%" ->  
      let x = Sedlexing.Utf8.lexeme buf in
      let x = String.sub x 0 (String.length x -1) in
      DOUBLE ((float_of_string x) /. 100. )
```

I'm wondering if we could simplify that, e.g., by using `as` like in ocamllex.

PR #112, opened 2022-09-19
Support named capture groups

dyzsr:
Related: #5, #102


src/syntax/ppx_sedlex.ml
```diff
@@ -311,16 +458,37 @@ let regexp_of_pattern env =
...
-  and aux p =
+  and aux ?(allow_alias = false) p =
+    let loc = p.ppat_loc in
     (* interpret one pattern node *)
     match p.ppat_desc with
-      | Ppat_or (p1, p2) -> Sedlex.alt (aux p1) (aux p2)
+      | Ppat_or (p1, p2) ->
+          let r1, s1 = aux ~allow_alias p1 in
+          let r2, s2 = aux ~allow_alias p2 in
+          if not (StrLocSet.equal s1 s2) then begin
```
hhugo, line 468: Seems to be a mistake to use StrLocSet to check if the two side of or-pattern define the same aliases

dyzsr: Yes. I'll fix it later.

dyzsr:
Can anybody help add reviewers?

src/syntax/ppx_sedlex.ml
```diff
@@ -311,16 +457,37 @@ let regexp_of_pattern env =
       | _ ->
           err p.ppat_loc @@ "the " ^ name
           ^ " operator requires two arguments, like " ^ name ^ "(a,b)"
-  and aux p =
+  and aux ?(allow_alias = false) p =
```
hhugo, line 460: Make allow_alias non optionnal. It will help to carefully review all recursive calls.

src/syntax/ppx_sedlex.ml
```diff
@@ -379,36 +547,46 @@ let regexp_of_pattern env =
...
+                  StrLocSet.empty )
             | _ -> err p.ppat_loc "this pattern is not a valid regexp"
         end
       | Ppat_var { txt = x } -> begin
-          try StringMap.find x env
+          try (StringMap.find x env, StrLocSet.empty)
           with Not_found ->
             err p.ppat_loc (Printf.sprintf "unbound regexp %s" x)
         end
+      | Ppat_alias (p, ({ txt = x } as x_loc)) when allow_alias ->
```
hhugo, line 580: Can we give a better error  for `Ppat_alias _ when not allow_alias` ?

src/syntax/ppx_sedlex.ml
```diff
@@ -489,9 +667,13 @@ let mapper =
         let l, regexps' = sub#structure_with_regexps (!previous @ l) in
         let parts = List.map partition (get_partitions ()) in
         let tables = List.map table (get_tables ()) in
+        let funcs =
+          let loc = default_loc in
+          [%str let __sedlex_tl = function _ :: tl -> tl | _ -> assert false]
```
hhugo, line 672: Can we put this in the lib, similar to what we do for `Sedlexing.__private__next_int`

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +266,157 @@ let gen_recflag auto =
...
+      @@ e
+
 let gen_definition lexbuf l error =
   let loc = default_loc in
-  let brs = Array.of_list l in
-  let auto = Sedlex.compile (Array.map fst brs) in
+  let brs =
+    Array.of_list
+      (List.map
+         (fun ((r, s), e) -> ((r, List.of_seq (StrLocSet.to_seq s)), e))
```
hhugo, line 395: `StrLocSet.elements` ?

toots (commented):
I can't really comment on the implementation but, for the API, is there a benefit in exporting `offset`/`length` vs simply binding the resulting match string?

hhugo:
Could you give more details about the current implementation ? Why is it done this way ? Can't we update the position as we traverse the automation, instead of doing a second pass based on some trace ? Why is the trace processed in reverse order ?

hhugo:
There is a lot of room for optimization.

Offset tracking could be shared between bindings.

For example, in`(Plus 'a' as a), (Plus 'b'  as b)`, end of binding a equals start of binding b.

Some offset can be computer statically. In the regexp above, start of a is 0, end of b is the end of the lexeme.

Have you considered such optimizations ?

hhugo:
```
let rec token buf =
  match%sedlex buf with
    | (Plus "a" as a), (Plus "b" as b), "c" ->
        ignore a;
        ignore b;
        Printf.printf "Op %s - %s\n"
          (Sedlexing.Latin1.sub_lexeme buf (fst a) (snd a))
          (Sedlexing.Latin1.sub_lexeme buf (fst b) (snd b));
        token buf
    | eof -> print_endline "EOF"
    | _ -> failwith "Unexpected character"

let () =
  let lexbuf = Sedlexing.Latin1.from_string "aaaabbbc" in
  token lexbuf
```

Returns an unexpected result:
```
Op aaaab - bb                       
EOF
```

hhugo:
I think I would be easier if your trace was maintaining pairs of pos, instead of pos + len.

dyzsr:
> I think I would be easier if your trace was maintaining pairs of pos, instead of pos + len.

Thanks for your suggestions! Let me do some fixes.

dyzsr:
> I can't really comment on the implementation but, for the API, is there a benefit in exporting `offset`/`length` vs simply binding the resulting match string?

The bindings can be either `string`, `Uchar.t array` or utf8/utf16 strings, so I think it's better to export just offset/length and let the user select the `sub_lexeme` function.

toots:
> > I can't really comment on the implementation but, for the API, is there a benefit in exporting `offset`/`length` vs simply binding the resulting match string?
> 
> The bindings can be either `string`, `Uchar.t array` or utf8/utf16 strings, so I think it's better to export just offset/length and let the user decide the `sub_lexeme` function.

Makes sense, ty!

hhugo:
I would be useful to write test (as expect test (ppx_expect) maybe). The examples are currently not testing enough.

dyzsr:
> I would be useful to write test (as expect test (ppx_expect) maybe). The examples are currently not testing enough.

Sure, I'll refactor the test later.

> ```
> Op aaaab - bb                       
> EOF
> ```

Give me some time and I'm working on this.

dyzsr:
> ```
> Op aaaab - bb                       
> EOF
> ```

@hhugo This one is fixed. You may check the expect test result.

I will put more details about the implementation in next few days. (It's kind of difficult to explain it clearly).

hhugo:
Another one
```
let print_alias buf name (x,len) =
  Printf.printf "%s: %S (%d, %d)\n" name (Sedlexing.Latin1.sub_lexeme buf  x len) x len
let rec token buf =
  match%sedlex buf with
  | (((Star "a" as a), (Plus "b" as b)), "c") | (Plus "a" as b), (Plus "b" as a), ("d") ->
    print_alias buf "a" a;
    print_alias buf "b" b;
        token buf
    | eof -> print_endline "EOF"
    | _ -> failwith "Unexpected character"

let () =
  let lexbuf = Sedlexing.Latin1.from_string "aaaabbbbd" in
  token lexbuf
```
wrongly returns 
```
a: "aaaa" (0, 4)                    
b: "bbbb" (4, 4)
EOF
```

hhugo:
I suspect you will need to track more than a pair of pos for some alias.

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,195 @@ let gen_recflag auto =
...
+                 } ->
+              let lhs =
+                ppat_tuple ~loc
+                  [
+                    pint ~loc curr_state;
+                    pint ~loc curr_node;
+                    pint ~loc prev_state;
+                  ]
+              in
+              let guard = Some (gen_cset ~loc "__sedlex_code" char_set) in
```
hhugo, line 336: IIUC, the guard is only useful when that are duplicate triples `(curr_state, curr_node, prev_state)` in `trans`

dyzsr: Yeah. My mistake

dyzsr:
With the information of matched lexeme and DFA path, we have to restore the NFA path. Tracing back the NFA path, we know from the NFA node about when an alias starts and stops.

For example, 

>   | (((Star "a" as a), (Plus "b" as b)), "c") | (Plus "a" as b), (Plus "b" as a), ("d") ->

A possible (simplified) NFA would be

``` c
0 ->(eps) 1
0 ->(eps) 5
1 ->[a] 1
1 ->(eps) 2
2 ->[b] 3
3 ->(eps) 2
3 ->(eps) 4
4 ->[c] 10
5 ->[a] 6
6 ->(eps) 5
6 ->(eps) 7
7 ->[b] 8
8 ->(eps) 7
8 ->(eps) 9
9 ->[d] 10
```

where 0 is the initial node and 10 is the final node. A converted DFA could be

``` c
1: {0, 1, 2, 5}
2: {1, 2, 5, 6, 7}
3: {2, 3, 4, 7, 8, 9}
4: {10}

1 ->[a] 2
2 ->[a] 2
2 ->[b] 3
3 ->[b] 3
3 ->[c, d] 4
```

where 1 is the initial state and 4 is the final state.

>   let lexbuf = Sedlexing.Latin1.from_string "aaaabbbbd" in

With input `"aaaabbbbd"`, we got a DFA path `[4; 3; 3; 3; 3; 2; 2; 2; 2; 1]`. The final NDA node is unique so we know it's 10. Starting from state 4 and node 10, we now trace back the path. The previous NFA node can be decided by `(current_state, current_node, previous_state, char_set)`, therefore a transition case is in the form of `(current_state, current_node, previous_state, char_set) => previous_node`. What `Sedlex.compile_traces` returns is a list of such transition cases.

The transition cases for DFA state 4 and NFA node 10 can be

``` c
(4, 10, 3, 'c') => 4
(4, 10, 3, 'd') => 9
```

Note that `previous_node` can travel to `current_node` by first going through a real edge (`node.trans`), and then several epsilon edges (`node.eps`). They are not always adjacent.

And since the starting and stoping of an alias can only happen when travel through epsilon edges (see `Sedlex.alias`), we just generate actions when the transition go through any epsilon edges (see `Sedlex.trans_case`).

The last DFA state we trace back will always be the initial state (which is 1), however the NFA node we are currently in may not be 0. It can be in 1, 2, or 5, and will travel back to 0 by only epsilon edges. Therefore we generate actions for the initial state (see `Sedlex.final_case`).

dyzsr:
> Could you give more details about the current implementation ? Why is it done this way ? Can't we update positions as we traverse the automation, instead of doing a second pass based on some trace ? Why is the trace processed in reverse order ?

The reason why making a second pass in the reverse order is that, with a valid given DFA path, you can always find an NFA path from the final node back to the initial node. But from a front order, there are some paths that cannot lead to the final node within the required steps.

And in the second pass, we already know which branch we're in, so nodes and aliases from other branches can be ignored. If do it in one pass, I'm afraid we have to refactor a lot to record the states and actions.

src/syntax/ppx_sedlex.ml
hhugo: Based on the fix you've committed, I think you misunderstand my point.

The generated code look like 
```ocaml
           | (4, 9, 3) when (98 <= __sedlex_code) && (__sedlex_code <= 98) ->
               (__sedlex_aliases_stops.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 3 7 __sedlex_rest)
           | (0, 5, 0) when (97 <= __sedlex_code) && (__sedlex_code <= 97) ->
               (__sedlex_aliases_stops.(0) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 0 3 __sedlex_rest)
           | (4, 21, 4) when (98 <= __sedlex_code) && (__sedlex_code <= 98)
               ->
               (__sedlex_aliases_stops.(0) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 4 18 __sedlex_rest)
``` 
the `when` guards are only useful when tripes  (e.g. `(4, 9, 3)`) are not unique.

```ocaml
           | (4, 9, 3)  ->
               (__sedlex_aliases_stops.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 3 7 __sedlex_rest)
           | (0, 5, 0)  ->
               (__sedlex_aliases_stops.(0) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 0 3 __sedlex_rest)
           | (4, 21, 4)  ->
               (__sedlex_aliases_stops.(0) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 4 18 __sedlex_rest)
```

hhugo:
I think you should look at minimizing the quantity of code generated for trace fonctions.  My last example generate a trace function with 80 cases. 

Have you though about optimizing the number positions tracked ? You could have a single array of positions (instead of one for starts and one for ends) and have alias reference a pair of position. position could then more easily be shared between aliases.

src/syntax/ppx_sedlex.ml
hhugo: This comment still stand

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,208 @@ let gen_recflag auto =
...
+      let alias_indexes =
+        List.to_seq aliases
+        |> Seq.mapi (fun i { txt = alias } -> (alias, i))
+        |> StringMap.of_seq
+      in
+      let find_index alias = StringMap.find alias alias_indexes in
+      let result_arrays =
+        [
+          value_binding ~loc
+            ~pat:[%pat? __sedlex_aliases_starts]
```
hhugo, line 294: unifying `__sedlex_aliases_starts` and `__sedlex_aliases_stops` into a single `__sedlex_aliases_pos` could be a first step toward sharing more positions between aliases.

hhugo:
This PR starts to look good, thanks for all the changes.

- I think it would be nice to commit the post-processed version of your test (like you did before) with the corresponding dune stanza so that the file is kept up to date automatically. I think this really helps when reviewing changes.

- More comments in the code would be welcome. Maybe inline some of the explanation made in this thread inside the code.

- You haven't reacted to one of my comment regarding optimization.
> There is a lot of room for optimization.
> Offset tracking could be shared between bindings.
> For example, in(Plus 'a' as a), (Plus 'b' as b), end of binding a equals start of binding b.
> Some offset can be computer statically. In the regexp above, start of a is 0, end of b is the end of the lexeme.
> Have you considered such optimizations ?

ocamllex has some optimization:
```
rule token = parse
  | ("a"+ as a) ("b" + as b) "d" { a,b }
```
only stores 2 integers

```
rule token = parse
  | (['-' '+'] as sign) (['0'-'9']+ as a) { a,b }
```
Doesn't store anything.


- The change on `.ocamlformat` needs to be removed before the merge.

hhugo:
You should also update the `README.md` with some documentation

dyzsr:
@hhugo Thanks for your considerate suggestions! Let me do it one by one and start from merging the alias offsets.

> ocamllex has some optimization:
> 
> ```
> rule token = parse
>   | ("a"+ as a) ("b" + as b) "d" { a,b }
> ```
> 
> only stores 2 integers
> 
> ```
> rule token = parse
>   | (['-' '+'] as sign) (['0'-'9']+ as a) { a,b }
> ```
> 
> Doesn't store anything.

I'm not sure if I'm able to support optimizations in such static cases. I'll think about it.

Another thing yet to do is to stop tracking the path if the matching branches have no alias at all.

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,228 @@ let gen_recflag auto =
...
+            actions)
+        cases;
+      let counter = ref 0 in
+      let alias2offset = Hashtbl.create n in
+      let cases2offset = Hashtbl.create 31 in
+      Hashtbl.iter
+        (fun action offsets ->
+          try
+            let i = Hashtbl.find cases2offset offsets in
+            Hashtbl.replace cases2offset offsets i;
```
hhugo, line 294: This `Hashtbl.replace` doesn't seem to be doing anything.

src/syntax/sedlex.mli
```diff
@@ -22,4 +23,18 @@ val intersection : regexp -> regexp -> regexp option
...
    which matches the intersection set.  Otherwise returns [None]. *)
 
-val compile : regexp array -> ((Sedlex_cset.t * int) array * bool array) array
+type trans_case = {
+  curr_state : int;
+  curr_node : int;
+  prev_state : int;
+  prev_node : int;
+  char_set : Sedlex_cset.t;
+  actions : (string * bool) list;
```
hhugo, line 32: maybe declare a separate type for action ?
`type action = Start of string | End of string` ?

src/syntax/ppx_sedlex.ml
hhugo: This has now been addressed.

src/syntax/ppx_sedlex.ml
hhugo: This comment is not relevant anymore

src/syntax/ppx_sedlex.ml
hhugo: This has been applied

src/syntax/ppx_sedlex.ml
hhugo: This now looks good.

src/syntax/ppx_sedlex.ml
hhugo: This has been done.

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,228 @@ let gen_recflag auto =
...
+                               __sedlex_rest
+                         | _ -> assert false);
+                       __sedlex_offsets]]];
+      ]
+
+let gen_aliases lexbuf offsets i e = function
+  | [] -> e
+  | aliases ->
+      let loc = default_loc in
+      let _, action_offsets = Option.get offsets.(i) in
```
hhugo, line 438: instead of matching on `aliases` above, we could directly match on `offsets.(i)`
```ocaml
let gen_aliases lexbuf offsets i e aliases =
  match offsets.(i) with
  | None -> e
  | Some (_, action_offsets) ->
     let loc = default_loc in
     pexp_let ~loc Nonrecursive
        [
          value_binding ~loc
            ~pat:[%pat? __sedlex_offsets]
            ~expr:
              (appfun (trace_fun i) [evar ~loc lexbuf; [%expr __sedlex_path]]);
        ]
      @@ pexp_let ~loc Nonrecursive
           (List.map
             (fun { txt = alias; loc } ->
                let start = Hashtbl.find action_offsets (alias, true) in
                let stop = Hashtbl.find action_offsets (alias, false) in
               value_binding ~loc ~pat:(pvar ~loc alias)
                  ~expr:
                    [%expr
                      __sedlex_offsets.([%e eint ~loc start]),
                        __sedlex_offsets.([%e eint ~loc stop])
                        - __sedlex_offsets.([%e eint ~loc start])])
              aliases)
      @@ e
```

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,228 @@ let gen_recflag auto =
...
+        (fun i actions ->
+          List.iter
+            (fun action ->
+              try
+                let offsets = Hashtbl.find action2cases action in
+                Hashtbl.replace action2cases action (i :: offsets)
+              with Not_found -> Hashtbl.add action2cases action [i])
+            actions)
+        cases;
+      let counter = ref 0 in
```
hhugo, line 287: counter could be replace by `Hastbl.length alias2offset` I think

hhugo: ignore me

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,228 @@ let gen_recflag auto =
...
+
+let gen_trace lexbuf traces i = function
+  | None -> []
+  | Some (offsets_num, action_offsets) ->
+      let loc = default_loc in
+      let initial, trans, finals = traces.(i) in
+      let offset_array =
+        value_binding ~loc
+          ~pat:[%pat? __sedlex_offsets]
+          ~expr:(pexp_array ~loc (List.init offsets_num (fun _ -> [%expr 0])))
```
hhugo, line 324: I would use an invalid position here (e.g. '-1`) as it would make it more obvious if we ever encounter a bug where position are not updated.

src/syntax/sedlex.ml
```diff
@@ -117,6 +127,89 @@ let transition (state : state) =
...
+      (fun from_state j ->
+        Hashtbl.iter
+          (fun to_state i ->
+            List.iter
+              (fun from_node ->
+                try
+                  let node_j = Hashtbl.find nodes_idx from_node in
+                  let rec dfs cset actions to_node =
+                    let node_i = Hashtbl.find nodes_idx to_node in
+                    try ignore (Hashtbl.find cases (i, node_i, j, cset))
```
hhugo, line 173: `Hashtbl.mem`

src/syntax/sedlex.ml
```diff
@@ -117,6 +129,89 @@ let transition (state : state) =
...
+  prev_node : int;
+  char_set : Sedlex_cset.t;
+  actions : action list;
+}
+
+type final_case = { curr_node : int; actions : action list }
+
+let compile_traces states (start, final) =
+  let counter = ref 0 in
+  let nodes_idx = Hashtbl.create 31 in
```
hhugo, line 147: How are `nodes_idx` different from `node.id` ?

src/syntax/sedlex.ml
```diff
@@ -117,6 +129,89 @@ let transition (state : state) =
...
+          states)
+      states;
+    Hashtbl.to_seq_values cases |> List.of_seq
+  in
+  let final_cases =
+    let rec dfs actions cases node =
+      let i = Hashtbl.find nodes_idx node in
+      let actions = append_action actions node.action in
+      let cases =
+        if node.trans <> [] || node == final then
```
hhugo, line 205: Can you add a comment explaining these conditions ?

src/syntax/sedlex.ml
dyzsr: Right. I forgot there's already an id in node.

src/syntax/sedlex.ml
dyzsr: Will add it later.

src/syntax/ppx_sedlex.ml
```diff
@@ -311,16 +523,37 @@ let regexp_of_pattern env =
...
+  and aux allow_alias p =
+    let loc = p.ppat_loc in
     (* interpret one pattern node *)
     match p.ppat_desc with
-      | Ppat_or (p1, p2) -> Sedlex.alt (aux p1) (aux p2)
+      | Ppat_or (p1, p2) ->
+          let r1, s1 = aux allow_alias p1 in
+          let r2, s2 = aux allow_alias p2 in
+          if not (StrLocSet.equal s1 s2) then begin
+            let x =
```
hhugo, line 534: What about providing the full list of problematic bindings ?
(e.g `StrLocSet.(diff (union s1 s2) (inter s1 s2))`)

src/syntax/ppx_sedlex.ml
```diff
@@ -311,16 +523,37 @@ let regexp_of_pattern env =
...
+            ^ " must occur on both sides of this | pattern"
+          end;
+          (Sedlex.alt r1 r2, s1)
       | Ppat_tuple (p :: pl) ->
-          List.fold_left (fun r p -> Sedlex.seq r (aux p)) (aux p) pl
+          List.fold_left
+            (fun (r1, s1) p ->
+              let r2, s2 = aux allow_alias p in
+              if not (StrLocSet.disjoint s1 s2) then begin
+                let x = StrLocSet.choose (StrLocSet.inter s1 s2) in
```
hhugo, line 547: Same here, we could return the full list of duplicated bindings

src/syntax/sedlex.ml
```diff
@@ -117,6 +129,89 @@ let transition (state : state) =
...
+  let first_node = Hashtbl.find nodes_idx final in
+  let trans_cases =
+    let cases = Hashtbl.create 31 in
+    Hashtbl.iter
+      (fun from_state j ->
+        Hashtbl.iter
+          (fun to_state i ->
+            List.iter
+              (fun from_node ->
+                try
```
hhugo, line 171: is this try-with necessary ?

src/syntax/ppx_sedlex.ml
```diff
@@ -253,24 +264,225 @@ let gen_recflag auto =
...
+                   prev_node;
+                   char_set;
+                   actions;
+                 } ->
+              let lhs =
+                ppat_tuple ~loc
+                  [
+                    pint ~loc curr_state;
+                    pint ~loc curr_node;
+                    pint ~loc prev_state;
```
hhugo, line 358: is `prev_state` ever needed to choose the action ?
On a trivial example
```ocaml
let rec token buf =
  match%sedlex buf with
    | (Star "a" as a), (Plus "b" as b), "c" ->
        print_alias buf "a" a;
        print_alias buf "b" b;
        token buf
    | eof -> print_endline "EOF"
    | _ -> failwith "Unexpected character"
```
`curr_state` and `curr_node` are enough for choosing the action.
The following
```ocaml
          let __sedlex_code = Sedlexing.lexeme_code buf (__sedlex_pos - 1) in
          (match (__sedlex_curr_state, __sedlex_curr_node,
                   __sedlex_prev_state)
           with
           | (2, 6, 0) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 0 10 __sedlex_rest)
           | (0, 6, 2) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 2 10 __sedlex_rest)
           | (3, 6, 2) -> __sedlex_aux (__sedlex_pos - 1) 2 6 __sedlex_rest
           | (4, 1, 3) -> __sedlex_aux (__sedlex_pos - 1) 3 2 __sedlex_rest
           | (3, 2, 0) ->
               (__sedlex_offsets.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 0 6 __sedlex_rest)
           | (3, 6, 0) -> __sedlex_aux (__sedlex_pos - 1) 0 6 __sedlex_rest
           | (2, 6, 2) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 2 10 __sedlex_rest)
           | (3, 2, 2) ->
               (__sedlex_offsets.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 2 6 __sedlex_rest)
           | (3, 2, 3) ->
               (__sedlex_offsets.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 3 6 __sedlex_rest)
           | (3, 6, 3) -> __sedlex_aux (__sedlex_pos - 1) 3 6 __sedlex_rest
           | (0, 10, 2) -> __sedlex_aux (__sedlex_pos - 1) 2 10 __sedlex_rest
           | (0, 6, 0) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) 0 10 __sedlex_rest)
           | (2, 10, 2) -> __sedlex_aux (__sedlex_pos - 1) 2 10 __sedlex_rest
           | (0, 10, 0) -> __sedlex_aux (__sedlex_pos - 1) 0 10 __sedlex_rest
           | (2, 10, 0) -> __sedlex_aux (__sedlex_pos - 1) 0 10 __sedlex_rest
           | _ -> assert false) in
```

can be rewritten into  

```ocaml
          let __sedlex_code = Sedlexing.lexeme_code buf (__sedlex_pos - 1) in
          (match (__sedlex_curr_state, __sedlex_curr_node) with
           | (2, 6) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 10
                  __sedlex_rest)
           | (0, 6) ->
               (__sedlex_offsets.(2) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 10
                  __sedlex_rest)
           | (3, 6) ->
               __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 6
                 __sedlex_rest
           | (4, 1) ->
               __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 2
                 __sedlex_rest
           | (3, 2) ->
               (__sedlex_offsets.(1) <- __sedlex_pos;
                __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 6
                  __sedlex_rest)
           | (0, 10) ->
               __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 10
                 __sedlex_rest
           | (2, 10) ->
               __sedlex_aux (__sedlex_pos - 1) __sedlex_prev_state 10
                 __sedlex_rest
           | _ -> assert false) in
```
if prev_state is sometime useful, we could include it inside a guard only when needed.

src/syntax/ppx_sedlex.ml
hhugo: This is done

src/syntax/sedlex.ml
hhugo: This is done

src/syntax/sedlex.ml
hhugo: Addressed

src/syntax/ppx_sedlex.ml
dyzsr: The `prev_state` is still useful. It's a good idea to make it a guard.

src/syntax/sedlex.ml
dyzsr: It caused some problem when I removed `nodes_idx`. Now it's fixed.

src/syntax/ppx_sedlex.ml
dyzsr: I think it's OK to report one location at a time, just like the OCaml compiler behaves.

src/syntax/ppx_sedlex.ml
dyzsr: I made a test and found that `prev_state` guards show up frequently, so I decide not to change it.

src/syntax/sedlex.ml
hhugo: Can you elaborate on the issue and explain what are the irrelevant nodes ?

pmetzger:
FYI, if one really wants to support capture groups correctly: https://re2c.org/#papers has several must-read papers.

dyzsr:
@pmetzger Thanks for the link!

src/syntax/ppx_sedlex.ml
```diff
@@ -311,16 +567,37 @@ let regexp_of_pattern env =
...
+    let loc = p.ppat_loc in
     (* interpret one pattern node *)
     match p.ppat_desc with
-      | Ppat_or (p1, p2) -> Sedlex.alt (aux p1) (aux p2)
+      | Ppat_or (p1, p2) ->
+          let r1, s1 = aux allow_alias p1 in
+          let r2, s2 = aux allow_alias p2 in
+          if not (StrLocSet.equal s1 s2) then begin
+            let x =
+              try StrLocSet.choose (StrLocSet.diff s1 s2)
```
hhugo, line 579: I would avoid the exception based selection. and use
`StrLocSet.choose  StrLocSet.(diff (union s1 s2) (inter s1 s2))`

hhugo:
> I think it would be nice to commit the post-processed version of your test (like you did before) with the corresponding dune stanza so that the file is kept up to date automatically. I think this really helps when reviewing changes.

I've added this in https://github.com/hhugo/sedlex/commit/0284bfd961d3c2c4bcda759fb0e03f68fb7e84ee

toots:
Hey there! Do we want to try & come to a consensus on wether this could be merged?

pmetzger:
@toots I think there's at least one code review comment above that hasn't yet been addressed? Generally I think we should merge though?

hhugo:
I currently can't close/resolve threads. There are still a bunch unaddressed ones.

hhugo:
replaced by #175

PR #117, opened 2023-02-17
Fix directly nested sedlex matches

smuenzel:
Currently, the code pattern
```
  match%sedlex lexbuf with
  ...
  | _ ->
    match%sedlex lexbuf with
    | _ ->
```
Does not compile, because the match subexpressions are transformed with `super#expression` rather than `this#expression`.


hhugo:
looks good

toots:
Thanks!

PR #118, opened 2023-02-17
Preserve location of lexbuf

### LINKED ISSUE #19: Fix location on lexbuf
E.g.:

```
let x y =
  match%sedlex lexbuf with _ -> assert false
```

produces this error:

```
File "_none_", line 1:
Error: Unbound value lexbuf
```

which is not very helpful.

hhugo:
fix #19 

before
```
$ dune build
File "_none_", line 1:                
Error: This expression has type int but an expression was expected of type
         Sedlexing.lexbuf
```

after 
```
$ dune build
File "examples/repeat.ml", line 4, characters 15-17:
4 |   match%sedlex rr with
                   ^^
Error: This expression has type int but an expression was expected of type
         Sedlexing.lexbuf
```

### DIFF
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -202,7 +202,7 @@ let call_state lexbuf auto state =
       | None -> assert false)
   else appfun (state_fun state) [lexbuf]
 
-let gen_state lexbuf auto i (trans, final) =
+let gen_state (lexbuf_name, lexbuf) auto i (trans, final) =
   let loc = default_loc in
   let partition = Array.map fst trans in
   let cases =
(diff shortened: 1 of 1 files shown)


src/syntax/ppx_sedlex.ml
```diff
@@ -211,22 +211,26 @@ let gen_state lexbuf auto i (trans, final) =
...
-            ~rhs:[%expr Sedlexing.backtrack [%e evar ~loc lexbuf]];
+            ~rhs:[%expr Sedlexing.backtrack [%e lexbuf]];
         ])
   in
   let ret body =
+    let lhs =
+      match lexbuf with
+        | { pexp_desc = Pexp_ident { txt = Lident ident; _ }; pexp_loc } ->
+            pvar ~loc:pexp_loc ident
+        | _ -> assert false
```
smuenzel, line 228: You could use the `Longident.t loc` for the location (and get it from `method! expression` as before).
That way you avoid the `assert false`

hhugo: I still need the string inside the simple `ident` to construct the pattern

smuenzel: What about https://github.com/hhugo/sedlex/compare/fixloc...smuenzel:sedlex:fixloc-explore ?
Shorter and we encode our invariants in the type system, rather than in `assert false`

smuenzel:
Other than the small comment I think this is good to go.

smuenzel (approved):

src/syntax/ppx_sedlex.ml
hhugo: looks good, I've cherry-pick you commit

hhugo:
@toots, this one looks ready for a merge, whenever you have time.

toots:
Looks great, thank you.

PR #122, opened 2023-02-17
Explicit stdlib in generated code

### LINKED ISSUE #115: Incompatible with -open flag
With a dune file like this:
```
(library
  (name foobar)
  (libraries sedlex core)
  (preprocess (pps sedlex.ppx)))
```

and a source file like this:
```ocaml
open! Core

let foobar buf =
  match%sedlex buf with
  | "{foo}" -> ()
  | "{bar}" -> ()
  | _ -> ()
;;
```

The code compiles fine. But, if you try to remove `open! Core` by doing `-open Core` in the dune file like so:
```
(library
  (name foobar)
  (libraries sedlex core)
  (flags -open Core)
  (preprocess (pps sedlex.ppx)))
```

The code will no longer compile. The reason is that the ppx generates some code at the top of the file which assumes that you have the OCaml stdlib in scope. If `open! Core` is included, it is done at the end of this block. This explains the reason why the first example works. But if you change it to `(flags -open Core)`, then the open actually happens at the top of the file, and generated ppx code is no longer referencing the OCaml stdlib.

### Proposal 1)
Stdlib modules should be prefixed by Stdlib. For example: `Stdlib.Char.code`.

### Proposal 2)
Wrap the generated code in a module, and open Stdlib in there, then open the module:
```ocaml
module Super_secret_private_module_for_sedlex = struct
  open! Stdlib

  (* generated code goes here *)
end
open Super_secret_private_module_for_sedlex
(* rest of the code goes here *)
```

### Proposal 3)
Doesn't OCaml support anonymous modules now? Maybe we could do 2) without having to name the module:
```ocaml
open struct
  open! Stdlib
  (* generated code goes here *)
end
(* rest of the code goes here *)
```

hhugo:
fix #115

### DIFF
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -177,7 +177,11 @@ let partition (name, p) =
           Char.code (String.unsafe_get [%e evar ~loc (table_name t)] [%e c]) - 1]
   in
   let body = gen_tree (simplify_decision_tree (decision_table p)) in
-  glb_value name [%expr fun c -> [%e body]]
+  glb_value name
+    [%expr
+      fun c ->
+        let open! Stdlib in
+        [%e body]]
 
 (* Code generation for the automata *)
 


toots (approved):
Looks great thanks!

PR #124, opened 2023-02-20
Don't use gen to consume channels

### LINKED ISSUE #45: Interactive mode

hhugo:
fix #45
replace #77


hhugo:
Reading 100_000_000 zero bytes from /dev/zero 
before: 2.44
after: 1.23

src/lib/sedlexing.ml
```diff
@@ -216,6 +212,38 @@ let with_tokenizer lexer' lexbuf =
...
+    else if can_refill then (
+      let len = t.len - t.pos in
+      if len > 0 then Bytes.blit t.b t.pos t.b 0 len;
+      t.pos <- 0;
+      let read = input t.ic t.b len (Bytes.length t.b - len) in
+      if read = 0 then raise Missing_input
+      else (
+        t.len <- len + read;
+        ensure t ~can_refill n))
+    else raise Missing_input
```
toots, line 239: Is this gonna work if `n` is larger than `len`? You should probably raise an exception if that's not the case.

src/lib/sedlexing.ml
```diff
@@ -512,7 +590,45 @@ module Utf16 = struct
...
+                      match (Char.code c1, Char.code c2) with
+                        | 0xff, 0xfe -> Little_endian
+                        | _ -> Big_endian
+                    in
+                    bo := Some o;
+                    o
+            in
+            let c = Helper.number_of_char_pair o c1 c2 in
+            Chan.advance t 2;
+            Uchar.of_int c
```
toots, line 620: Looks like a good portion of this logic could be factored out to make it more readable. Something along these lines:

```
      let rec loop ~fetch i =
        if i = len then i
        else (
          match
            let can_refill = i = 0 in
            Chan.ensure t ~can_refill 1;
            let (c, n) = fetch t in
            Chan.advance t n;
            Uchar.of_char c
        )
      in
```
This would make the code more readable for anyone not familiar with it.

toots (commented):
Some initial comments. Overall it looks like a worthwhile optimization with a nice fix for one of my issues! However, this also adds a lot of specialized code. I think some factorization as mentioned would be nice and also perhaps some tests?

src/lib/sedlexing.ml
hhugo: in practice, `n` is 1,2,3 or 4. I can add an assert if you prefer.
I'm not sure what `len` your are referring to.

toots: Yes I can see that but for users not familiar with the code, this won't be clear. Adding an assert will make it easier to proof read for any future contributor.

toots: Also, in this case why not set the length to 4 instead of parametrizing it?

hhugo: The length of the buffer is (much) bigger than four, allowing to read many bytes at once.

hhugo: I've added some check

src/lib/sedlexing.ml
```diff
@@ -250,21 +298,23 @@ module Utf8 = struct
...
+          width.(i) <- 2
+        done;
+        for i = 224 to 239 do
+          width.(i) <- 3
+        done;
+        for i = 240 to 247 do
+          width.(i) <- 4
+        done
+      in
+      fun c -> width.(Char.code c)
```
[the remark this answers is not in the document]
hhugo: I've cleaned up a bit more

src/lib/sedlexing.ml
hhugo: I've done some more refactoring

src/lib/sedlexing.ml
```diff
@@ -299,22 +311,16 @@ module Utf8 = struct
...
-          width.(i) <- 4
-        done
+      let a =
+        Array.init 256 (fun x ->
+            match Char.chr x with
+              | '\000' .. '\127' -> 1
+              | '\192' .. '\223' -> 2
+              | '\224' .. '\239' -> 3
+              | '\240' .. '\247' -> 4
+              | _ -> -1)
```
toots, line 321: 🔥

toots: Actually now that I'm thinking about it, why not simply a function? 😅

```ocaml
let width = function
  | '\000' .. '\127' -> 1
  | '\192' .. '\223' -> 2
  | '\224' .. '\239' -> 3
  | '\240' .. '\247' -> 4
  | _ -> assert false

**Review by toots — APPROVED** [review]
This looks great to me. Adding an approval to help move things forward but I think the following should be considered:

* Give others some time to chime in
* Maybe some tests? Make sure there's no regression and prevent future one.
* Address my last minor comment.

Thanks!

File: src/lib/sedlexing.ml (hunk 4 above)
Comment (hhugo) [reply src/lib/sedlexing.ml]: I wanted to preserve the existing ideal to not affect perf but I don't think it should matter much. I've pushed the change.
We need to keep the `-1` for malformed input.

File: src/lib/sedlexing.ml [hunk 5]
```diff
@@ -216,6 +212,50 @@ let with_tokenizer lexer' lexbuf =
...
+
+  let min_buffer_size = 64
+
+  let create ic len : t =
+    let len = max len min_buffer_size in
+    { b = Bytes.create len; ic; len = 0; pos = 0 }
+
+  let available (t : t) = t.len - t.pos
+
+  let rec ensure (t : t) ~can_refill n =
```
smuenzel, line 233: `ensure_bytes_available`?

src/lib/sedlexing.ml
```diff
@@ -228,7 +268,28 @@ module Latin1 = struct
       finished = true;
     }
 
-  let from_channel ic = from_gen (gen_of_channel ic)
+  let from_channel ic =
```
smuenzel, line 271: All the `from_channel` functions are pretty similar, can we factor out the common logic?

Such as
```
let make_from_channel ~buffer_multiplier ~min_bytes_per_uchar ~read_uchar_from_channel =
  fun ic ->
    let t = Chan.create ic (chunk_size * buffer_multiplier) in
    let bo = ref opt_bo in
    let refill buf pos len =
      let rec loop i =
        if i = len then i
        else (
          match
            (* we refill our bytes buffer only if we haven't refilled any uchar yet. *)
            let can_refill = i = 0 in
            Chan.ensure t ~can_refill min_bytes_per_uchar;
						read_uchar_from_channel t
          with
            | c ->
                buf.(pos + i) <- c;
                loop (i + 1)
            | exception Chan.Missing_input ->
                if i = 0 && Chan.available t > 0 then raise MalFormed;
                i)
      in
      loop 0
    in
    create refill
```

src/lib/sedlexing.ml
```diff
@@ -250,21 +311,12 @@ module Utf8 = struct
...
-      done;
-      for i = 240 to 247 do
-        width.(i) <- 4
-      done
+    let width = function
+      | '\000' .. '\127' -> 1
+      | '\192' .. '\223' -> 2
+      | '\224' .. '\239' -> 3
+      | '\240' .. '\247' -> 4
+      | _ -> -1
```
smuenzel, line 319: Why not `raise Malformed` here directly? Performance?

src/lib/sedlexing.ml
```diff
@@ -436,23 +518,24 @@ module Utf16 = struct
...
       else raise MalFormed
 
+    let get_bo bo c1 c2 =
+      match !bo with
+        | Some o -> o
+        | None ->
+            let o =
+              match (Char.code c1, Char.code c2) with
+                | 0xff, 0xfe -> Little_endian
+                | _ -> Big_endian
```
smuenzel, line 528: Minor comment: Even though the unicode standard says that this is correct behaviour, wikipedia says it's not really true in practice, people use little ending without a bom

smuenzel (commented):
Seems pretty straightforward, I left some small comments.

src/lib/sedlexing.ml
hhugo: Why not ! I've pushed the change.

src/lib/sedlexing.ml
hhugo: I've made the change

src/lib/sedlexing.ml
hhugo: done

src/lib/sedlexing.ml
hhugo: Not sure what to do with that, maybe create a new issue for this ?

hhugo:
I believe truncated utf8/utf16 were not properly recognized as malformed. My last commit https://github.com/ocaml-community/sedlex/pull/124/commits/5851233b8943d1c9ebe94192f974d452654eb8ee tries to fix this.

smuenzel (approved):

hhugo:
I've added some tests push fixes

hhugo:
I've rebased, squashed commits again, move test first so that we can see the effect of the change in the test.

toots (approved):
This looks great thanks for the hard work and integration with the CI. I'm ready to merge it just one more comment.

src/lib/sedlexing.ml
smuenzel: yeah, it's the same as the existing behaviour -- my comment was just for future reference

PR #126, opened 2023-02-24
Raise Malformed instead of Invalid_arg

### LINKED ISSUE #91: Instead of raising `Sedlexing.MalFormed` a `Invalid_argument` is triggered
When trying to parse certain malformed unicode characters, Sedlexing calls `Uchar.of_int` without sufficient verification that the codepoint is well-formed. This results in `Uchar.of_int` triggering a `Stdlib.Invalid_argument s` where `s` is some error message string. Instead, `MalFormed` should be raised, which would allow clean handling of this particular exception from the sedlex user side instead of a catch-all treatment of `Invalid_argument`.

An example file that contains such a malformed codepoint triggering this error can be found at
https://github.com/Andromedans/andromeda/blob/c45d500b3650a6df30d5a1ef09e2d677aeacc190/tests/malformed-unicode.m31

hhugo:
fix #91

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -5,6 +5,13 @@
 exception InvalidCodepoint of int
 exception MalFormed
 
+module Uchar = struct
+  include Uchar
+
+  let of_int x =
+    if Uchar.is_valid x then Uchar.unsafe_of_int x else raise MalFormed
+end
+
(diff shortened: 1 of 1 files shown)


hhugo:
A changelog entry would be nice as people have been working around this behavior. Ill be away from keyboard. Would you be able to add one.

toots:
Done!

PR #127, opened 2023-02-25
Add utf8 support for string literal

### LINKED ISSUE #90: Matching a unicode character without codepoint
Currently,
`
match%sedlex lexbuf with
 | "ρ" -> ..
`
does not match, although
`
match%sedlex lexbuf with
 | math -> if Sedlexing.Utf8.lexeme lexbuf = "ρ" then ..
`
does.

Is there any way of making the first variant work, without having to replace "ρ" with its underlying codepoint, so that I can use it as part of a more complex regexp?

hhugo:
The PR adds support for utf-8 encoded string. It fixes #90.

This only work if we assume the source is encoded in utf8.

Note that OCaml 5.3 has the following entry in its changelog.

> - #11736, #12664: Support utf-8 encoded source files and latin-9 compatible
  identifiers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The actions can call functions from the Sedlexing module to extract
 
 Regular expressions are syntactically OCaml patterns:
 
-- `"...."` (string constant): recognize the specified string
+- `"...."` (string constant): recognize the specified string.
 - `'....'` (character constant) : recognize the specified character
 - `i` (integer constant) : recognize the specified codepoint
 - `'...' .. '...'`: character range
@@ -103,6 +103,9 @@ Regular expressions are syntactically OCaml patterns:
   and recognize the set of items in `R1` but not in `R2` ("subtract")
 - `Intersect (R1,R2)` : assume that `R` is a single-character length regexp (see
   below) and recognize the set of items which are in both `R1` and `R2`
+- `Utf8 R` : string literals inside R are assumed to be utf-8 encoded.
+- `Latin1 R` : string literals inside R are assumed to be latin1 encoded.
+- `Ascii R` : string literals inside R are assumed to be ascii encoded.
 - `lid` (lowercase identifier) : reference a named regexp (see below)
 
 A single-character length regexp is a regexp which does not contain (after
@@ -112,8 +115,9 @@ with a length different from one.
 
 
 Note:
- - The OCaml source is assumed to be encoded in Latin1 (for string
-   and character literals).
+ - The OCaml source is assumed to be encoded in UTF-8.
+ - Strings and chars litterals will be interpreted in ASCII unless otherwise
+   specified by the `Latin1`,`Ascii` and `Utf8` constructors in patterns.
 
 
 It is possible to define named regular expressions with the following
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -13,7 +13,6 @@ module Cset = Sedlex_cset
 (* Decision tree for partitions *)
 
 let default_loc = Location.none
-let lident_loc ~loc s = { loc; txt = lident s }
 
 type decision_tree =
   | Lte of int * decision_tree * decision_tree
@@ -286,14 +285,8 @@ let codepoint i =
     failwith (Printf.sprintf "Invalid Unicode code point: %i" i);
   i
 
-let regexp_for_char c = Sedlex.chars (Cset.singleton (Char.code c))
-
-let regexp_for_string s =
-  let rec aux n =
-    if n = String.length s then Sedlex.eps
-    else Sedlex.seq (regexp_for_char s.[n]) (aux (succ n))
-  in
-  aux 0
+let char c = Cset.singleton (Char.code c)
+let uchar c = Cset.singleton (Uchar.to_int c)
 
 let err loc fmt =
   Printf.ksprintf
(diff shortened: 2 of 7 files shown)


pmetzger:
So the documentation doesn't (yet!) officially say if we can assume source files are UTF-8, but I presume that's imminent.

hhugo:
https://github.com/ocaml/ocaml/pull/13668

toots:
Any interest on reviving and finishing this?

hhugo:
I've debased the PR. code review welcome.

src/syntax/ppx_sedlex.ml
```diff
@@ -382,11 +394,9 @@ let regexp_of_pattern env =
...
           match const with
             | Some (Pconst_string (s, _, _)) ->
-                let c = ref Cset.empty in
-                for i = 0 to String.length s - 1 do
-                  c := Cset.union !c (Cset.singleton (Char.code s.[i]))
-                done;
-                Sedlex.chars !c
+                let l = rev_csets_of_string ~loc:p.ppat_loc ~utf8 s in
+                let chars = List.fold_left Cset.union Cset.empty l in
+                Sedlex.chars chars
```
toots, line 399: Nice

src/syntax/utf8.ml
```diff
@@ -0,0 +1,73 @@
+let unsafe_byte s j = Char.code (String.unsafe_get s j)
+let malformed s j l = `Malformed (String.sub s j l)
+
+let width = function
+  | '\000' .. '\127' -> 1
+  | '\192' .. '\223' -> 2
+  | '\224' .. '\239' -> 3
+  | '\240' .. '\247' -> 4
+  | _ -> 0
```
toots, line 9: Do you need to raise here?

src/syntax/utf8.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let r_utf_8 s j l =
+  (* assert (0 <= j && 0 <= l && j + l <= String.length s); *)
+  let uchar c = `Uchar (Uchar.unsafe_of_int c) in
+  match l with
+    | 1 -> uchar (unsafe_byte s j)
+    | 2 ->
+        let b0 = unsafe_byte s j in
+        let b1 = unsafe_byte s (j + 1) in
+        if b1 lsr 6 != 0b10 then malformed s j l
```
toots, line 19: This should be re-using the new checks in https://github.com/ocaml-community/sedlex/pull/168 when they get merged

toots (approved):
This looks good! The new checks from https://github.com/ocaml-community/sedlex/pull/168 should be added here once it's merged (which I'll do today).

Also, you should document the new `Utf8` in `README.md`. 

Other than that, this seems good to go for me.

src/syntax/utf8.ml
hhugo: now using `Sedlexing.Utf8.Helper.width`

src/syntax/utf8.ml
hhugo: done

toots:
Looking good!

hhugo:
I think this PR mostly need documentation now

test/utf8.ml
```diff
@@ -0,0 +1,19 @@
+open Printf
+
+let next_tok buf =
+  let open Sedlexing.Utf8 in
+  match%sedlex buf with
+    | "a", Utf8 (Chars "+-×÷") -> sprintf "with Chars: %s" (lexeme buf)
+    | "b", Utf8 ("+" | "-" | "×" | "÷") ->
```
toots, line 7: Could be nice to have an explcit non-ascii character here to test all the way.

toots:
Are you considering adding documentation @hhugo ?

src/syntax/ppx_sedlex.ml
```diff
@@ -286,29 +285,49 @@ let codepoint i =
...
+        Utf8.fold
+          ~f:(fun acc _ x ->
+            match x with
+              | `Malformed _ -> err loc "Malformed utf-8 string"
+              | `Uchar c -> uchar c :: acc)
+          [] s
+    | Latin1 ->
+        let l = ref [] in
+        for i = 0 to String.length s - 1 do
+          l := char s.[i] :: !l
```
toots, line 308: According to https://en.wikipedia.org/wiki/ISO/IEC_8859-1, not all values are proper `latin1` characters.

test/utf8.ml
```diff
@@ -0,0 +1,19 @@
...
+
+let%expect_test _ =
+  Sedlexing.Utf8.from_string "a+" |> next_tok |> print_string;
+  [%expect {| with Chars: a+ |}];
+  Sedlexing.Utf8.from_string "a÷" |> next_tok |> print_string;
+  [%expect {| with Chars: a÷ |}];
+  Sedlexing.Utf8.from_string "b+" |> next_tok |> print_string;
+  [%expect {| with or_pattern: b+ |}];
+  Sedlexing.Utf8.from_string "b÷" |> next_tok |> print_string;
+  [%expect {| with or_pattern: b÷ |}]
```
toots, line 19: Should we have test for cases that do not work? And `latin1`/`ascii` tests?

toots (commented):
This is looking great, just minor comments and one about valid latin1 chars.

src/syntax/ppx_sedlex.ml
hhugo:
```
In 1990, the first version of [Unicode](https://en.wikipedia.org/wiki/Unicode) used the code points of ISO-8859-1 as the first 256 Unicode code points.
```

toots: ok!

test/utf8.ml
hhugo: I've added some tests for Latin1 and Ascii.

It's not super easy to tests for errors currently.

README.md
```diff
@@ -112,8 +115,9 @@ with a length different from one.
 
 
 Note:
- - The OCaml source is assumed to be encoded in Latin1 (for string
-   and character literals).
+ - The OCaml source is assumed to be encoded in UTF-8.
+ - Strings and chars litterals will be interpreted in ASCII unless otherwise
+   specified by the `Latin1`,`Ascii` and `Utf8` constructors in patterns.
```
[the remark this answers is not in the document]
hhugo: The old behavior was to expect latin1 but it seems some people were also feeding utf8 https://github.com/FStarLang/FStar/blob/3ff998c60bb0efe9925fc94e8fb8b785b9485af0/src/ml/FStarC_Parser_LexFStar.ml#L530 or using the 0x80..0xff range weirdly https://github.com/janestreet/ppx_css/blob/3837ab33ee0c4635826db2c746e9da163a80a360/vendor/css_parser/src/lexer.ml#L93. It seems that theses are all incorrectly using sedlex but I can give this a deeper look next week

PR #130, opened 2023-06-01
Count lines in all cases.

toots:
It took me a while to track this down: we are not counting position lines when parsing from a string.

Is there any reason for that? This seems pretty counter-intuitive and I imagine most users would trip on that, expecting lines to be counted properly in all cases.

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -71,11 +71,7 @@ let dummy_uchar = Uchar.of_int 0
 let nl_uchar = Uchar.of_int 10
 
 let create refill =
-  {
-    empty_lexbuf with
-    refill;
-    buf = Array.make chunk_size dummy_uchar;
-  }
+  { empty_lexbuf with refill; buf = Array.make chunk_size dummy_uchar }
 
 let set_position lexbuf position =
   lexbuf.offset <- position.Lexing.pos_cnum - lexbuf.pos;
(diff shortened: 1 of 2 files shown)


src/lib/sedlexing.ml
```diff
@@ -133,9 +128,7 @@ let refill lexbuf =
   let n = lexbuf.refill lexbuf.buf lexbuf.pos chunk_size in
   if n = 0 then lexbuf.finished <- true else lexbuf.len <- lexbuf.len + n
 
-let new_line lexbuf =
-  if lexbuf.curr_line != 0 then lexbuf.curr_line <- lexbuf.curr_line + 1;
-  lexbuf.curr_bol <- lexbuf.pos + lexbuf.offset
+let new_line lexbuf = lexbuf.curr_line <- lexbuf.curr_line + 1
```
alainfrisch, line 131: It has been years since I looked/used sedlex, so perhaps I'm missing something obvious, but why did you drop `lexbuf.curr_bol <- lexbuf.pos + lexbuf.offset`?

toots: My bad, adding it back.

hhugo:
You need to update comments and doc-comments in the ml and MLI files.

toots:
Sounds good I was sending this first to see if there was a consensus on the changes. I'll wrap this up, thanks!

toots:
Let's see about https://github.com/ocaml-community/sedlex/pull/131 first and then I'll rebase this one to use the changelog entry from it. Also, https://github.com/ocaml-community/sedlex/pull/131 implies this on all constructors I believe so this PR would just become a cleanup basically.

hhugo:
You can now rebase the PR and update doc-comments :)

toots:
Yeah, it became pretty trivial after all 😅

PR #131, opened 2023-06-01
Use from_gen to parse utf* strings.

toots:
This PR optimizes memory and CPU usage but using the `from_gen` API to parse strings in `UTF8` and `UTF16` modules.


hhugo:
Some tests be nice.
Also, what the performance impact of your change for valid input ?

toots:
Here's a test program:

```ocaml
let read_whole_file filename =
  let ch = open_in_bin filename in
  let s = really_input_string ch (in_channel_length ch) in
  close_in ch;
  s

let () =
  let s = read_whole_file "/tmp/medialib.json" in
  let lexbuf = Sedlexing.Utf8.from_string s in
  ignore lexbuf
```

Tested on a `4.8M` json file:

Before the changes:
```shell
% time ./sedbench
./sedbench  0.06s user 0.01s system 96% cpu 0.080 total
```

After the changes:
```shell
% time ./sedbench
./sedbench  0.00s user 0.00s system 74% cpu 0.004 total
```

There seems to be a difference between cold and warm during testing so these are the result of the second run in each case and it was confirmed that subsequent runs where in the same range.

Overall, it's about 20x faster to use `from_gen` to _initialize_ the lexbuf!

Changelog entry has been added, this PR should be read for merge now!

hhugo:
Your benchmark doesn't lex anything, it just create the lexbuf. I would like to see the result that include lexing the whole content.

toots:
> Your benchmark doesn't lex anything, it just create the lexbuf. I would like to see the result that include lexing the whole content.

Since the patch is focused on the lexbuf creation, I would expect the gains to be on the lexbuf creation and anything after that to be additively static.

Indeed:

Without patch:
```
% time ./_build/default/bench.exe
DONE
./_build/default/bench.exe  0.09s user 0.01s system 95% cpu 0.112 total
```

With patch:
```
❯ time ./_build/default/bench.exe
DONE
./_build/default/bench.exe  0.00s user 0.00s system 80% cpu 0.008 total
```

<details>
  <summary>Benchmark code</summary>

```ocaml
open Sedlexing

type token =
  | EOF
  | LBRACKET
  | RBRACKET
  | COLON
  | COMMA
  | MINUS
  | PLUS
  | LSQUARE
  | RSQUARE
  | STRING of string
  | BOOL of bool
  | NUMBER of string

exception LexError of Lexing.position * string

let digit = [%sedlex.regexp? '0' .. '9']
let number = [%sedlex.regexp? Plus digit]
let blank = [%sedlex.regexp? ' ' | '\t']
let newline = [%sedlex.regexp? '\r' | '\n' | "\r\n"]
let any_blank = [%sedlex.regexp? blank | newline]
let letter = [%sedlex.regexp? 'a' .. 'z' | 'A' .. 'Z']
let decimal_ascii = [%sedlex.regexp? Plus '0' .. '9']
let octal_ascii = [%sedlex.regexp? "0o", Plus '0' .. '7']

let hex_ascii =
  [%sedlex.regexp? "0x", Plus ('0' .. '9' | 'a' .. 'f' | 'A' .. 'F')]

let rec nom buf = match%sedlex buf with Plus any_blank -> nom buf | _ -> ()

let string buf =
  let buffer = Buffer.create 10 in
  let rec read_string buf =
    match%sedlex buf with
      | {|\"|} ->
          Buffer.add_char buffer '"';
          read_string buf
      | '"' -> STRING (Buffer.contents buffer)
      | Star (Compl '"') ->
          Buffer.add_string buffer (Utf8.lexeme buf);
          read_string buf
      | _ -> assert false
  in
  read_string buf

let digit_value c =
  let open Stdlib in
  match c with
    | 'a' .. 'f' -> 10 + Char.code c - Char.code 'a'
    | 'A' .. 'F' -> 10 + Char.code c - Char.code 'A'
    | '0' .. '9' -> Char.code c - Char.code '0'
    | _ -> assert false

let num_value buffer ~base ~first =
  let buf = Utf8.lexeme buffer in
  let c = ref 0 in
  for i = first to String.length buf - 1 do
    let v = digit_value buf.[i] in
    assert (v < base);
    c := (base * !c) + v
  done;
  ignore !c;
  buf

let token buf =
  nom buf;
  match%sedlex buf with
    | eof -> EOF
    | "" -> EOF
    | '-' -> MINUS
    | '+' -> PLUS
    | '"' -> string buf
    | ':' -> COLON
    | '[' -> LSQUARE
    | ']' -> RSQUARE
    | '{' -> LBRACKET
    | '}' -> RBRACKET
    | ',' -> COMMA
    | "true" -> BOOL true
    | "false" -> BOOL false
    | hex_ascii ->
        let number = num_value ~base:16 ~first:2 buf in
        NUMBER number
    | octal_ascii ->
        let number = num_value ~base:8 ~first:2 buf in
        NUMBER number
    | decimal_ascii ->
        let number = num_value ~base:10 ~first:0 buf in
        NUMBER number
    | _ ->
        let position = fst @@ lexing_positions buf in
        let tok = Utf8.lexeme buf in
        raise
        @@ LexError (position, Printf.sprintf "unexpected character %S" tok)

let rec read_all buf =
  match token buf with
    | EOF -> Printf.eprintf "DONE\n"
    | tok ->
        ignore tok;
        read_all buf

let read_whole_file filename =
  let ch = open_in_bin filename in
  let s = really_input_string ch (in_channel_length ch) in
  close_in ch;
  s

let () =
  let s = read_whole_file "/tmp/medialib.json" in
  let lexbuf = Sedlexing.Utf8.from_string s in
  read_all lexbuf
```
</details>

toots:
@hhugo I think this one is a slam dunk frankly. Shall we merge it?

hhugo:
I'm happy for this to be merged eventually but I still think we should add some tests.

Also a few comments:

- I don't think the PR is enough to really address the issue mentioned in the description. With the proposed implementation `Malformed` will be raised during refill, which will not allow you to have the location of the error.

- Why fix utf8 and not utf16 ?

- `Sedlexing.from_int_array` has the same issue..

src/lib/sedlexing.ml
```diff
@@ -471,7 +450,15 @@ module Utf8 = struct
   let from_gen s =
     create (fill_buf_from_gen (fun id -> id) (Helper.gen_from_char_gen s))
 
-  let from_string s = from_int_array (Helper.to_int_array s 0 (String.length s))
+  let from_string s =
+    let gen =
```
hhugo, line 454: This can be implemented with `Gen.init ~limit:(String.length s) (fun i -> String.get s i)`

toots: Done thanks.

toots:
> I'm happy for this to be merged eventually but I still think we should add some tests.

There are a bunch of tests in `test/basic.ml` that seem designed to catch regression on all sort of lexbuf creation functions for `latin1`, `utf8` and `utf16`. What additional tests do you have in mind?

> Also a few comments:
> 
> * I don't think the PR is enough to really address the issue mentioned in the description. With the proposed implementation `Malformed` will be raised during refill, which will not allow you to have the location of the error.

Description clarified.

> * Why fix utf8 and not utf16 ?

Done, thanks for pointing out.

> * `Sedlexing.from_int_array` has the same issue..

I'm not sure that I see the issue:

```ocaml
let from_int_array a =
  let len = Array.length a in
  {
    empty_lexbuf with
    buf = Array.init len (fun i -> Uchar.of_int a.(i));
    len;
    finished = true;
  }
```

The original array doesn't seem copied here to me. And it's already in memory at call time. Pretty similar to having the string in memory at call time and peeling off of it during lexing.

hhugo:
The issue with `from_int_array` is that `Uchar.of_int` can raise if the integer is not a valid codepoint.

hhugo:
> There are a bunch of tests in test/basic.ml that seem designed to catch regression on all sort of lexbuf creation functions for latin1, utf8 and utf16. What additional tests do you have in mind?

I thing we should add tests where we partially lex the content with a malformed content on the part of the input that is not lexed.
(e.g. lexing an integer and a space from `123 \255\255`)

toots:
While I think these are nice, this is also out of the scope of this PR. If you prefer I can remove all reference to `MalFormed` and merge this strictly as a performance optimisation, which is what I'm looking at with these changes. Better handling of `MalFormed` should be handled in your refactor.

hhugo:
> With these changes, MalFormed errors might still be thrown while filling up the initial buffer when calling from_string but at least some errors that appears later in the string content will properly be raised during lexing.

- I don't think the buffer it filled during lexbuf creation.
- Malformed error raised during lexing will be done while refilling the buffer. The location wont be accurate.

Have you seen my reply on from_int_array ?

> The issue with `from_int_array` is that `Uchar.of_int` can raise if the integer is not a valid codepoint.

toots:
Hi. Sorry I got busy.

I would like to merge this PR as a _performance improvement_. I have update the PR description. Do you have any pending concern about performances and/or regressions with it?

.ocamlformat-ignore
```diff
@@ -0,0 +1 @@
+src/syntax/unicode.ml
```
hhugo, line 1: This should not be necessary. There is already an attribute in the ml file to disable ocamlformat. I see there is a small diff, it should be easy to fix the generator to remove the diff.

hhugo (commented):
Looks ok. just minor comments.

toots:
Ok great, thanks.

Merging now, will look at a release soon.

PR #132, opened 2023-06-06
Restore compat with ocaml 4.8

hhugo:
sedlex is still compatible with OCaml 4.08.

There is however no version of ppx_expect compatible with both ocaml 4.08 and  ppxlib 0.26.
I've disable the expect tests on ocaml version older than 4.14

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,10 @@ jobs:
           - 4.14.x
         skip-test:
           - false
+        include:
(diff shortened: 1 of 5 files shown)


toots:
Great thank you. Would you mind adding a new Changelog entry? We should do a quick release after that. I'll follow-up on my PR from your last comment shortly.

toots (approved):

hhugo:
Done

PR #134, opened 2023-06-07
Uniform handling of inputs

### LINKED ISSUE #133: lexbuf should handle invalid unicode input the same way regardless of how it is created.
Precisely, the following should behave the same when facing invalid input.
- Utf8.from_string
- Utf8.from_gen
- Utf8.from_int_array
- from_int_array
- from_gen

In order for this to work, we need to raise `MalFormed` only when all previous uchar have been consumed already.

hhugo:
Attempt to fix #133

- [x] make sure the code compile with OCaml 4.08
- [x] update tests
- [ ] run benchmarks

In practice, the code also removes some unicode logic duplication.

### DIFF
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -450,7 +450,7 @@ module Utf8 = struct
         Buffer.add_char b (Char.chr (0x80 lor (p land 0x3f))))
       else raise MalFormed
 
-    let from_uchar_array a apos len =
+    let string_of_uchar_array a apos len =
       let b = Buffer.create (len * 4) in
       let rec aux apos len =
         if len > 0 then (
@@ -478,7 +478,7 @@ module Utf8 = struct
   let from_string s = from_int_array (Helper.to_int_array s 0 (String.length s))
 
   let sub_lexeme lexbuf pos len =
-    Helper.from_uchar_array lexbuf.buf (lexbuf.start_pos + pos) len
+    Helper.string_of_uchar_array lexbuf.buf (lexbuf.start_pos + pos) len
 
   let lexeme lexbuf = sub_lexeme lexbuf 0 (lexbuf.pos - lexbuf.start_pos)
 end
@@ -575,7 +575,7 @@ module Utf16 = struct
         Buffer.add_char buf c3;
         Buffer.add_char buf c4)
 
-    let from_uchar_array bo a apos len bom =
+    let string_of_uchar_array bo a apos len bom =
       let b = Buffer.create ((len * 4) + 2) in
       (* +2 for the BOM *)
       if bom then store bo b 0xfeff;
@@ -619,7 +619,7 @@ module Utf16 = struct
     from_uchar_array a
 
   let sub_lexeme lb pos len bo bom =
-    Helper.from_uchar_array bo lb.buf (lb.start_pos + pos) len bom
+    Helper.string_of_uchar_array bo lb.buf (lb.start_pos + pos) len bom
 
   let lexeme lb bo bom = sub_lexeme lb 0 (lb.pos - lb.start_pos) bo bom
 end
--- a/src/lib/sedlexing.ml
+++ b/src/lib/sedlexing.ml
@@ -80,19 +80,21 @@ let set_position lexbuf position =
 
 let set_filename lexbuf fname = lexbuf.filename <- fname
 
-let fill_buf_from_gen f gen buf pos len =
-  let rec aux i =
-    if i >= len then len
-    else (
-      match gen () with
-        | Some c ->
-            buf.(pos + i) <- f c;
-            aux (i + 1)
-        | None -> i)
-  in
-  aux 0
+let fill_buf_from_gen gen =
+  let () = () in
+  fun buf pos len ->
+    let rec aux i =
+      if i >= len then len
+      else (
+        match gen () with
+          | Some c ->
+              buf.(pos + i) <- c;
+              aux (i + 1)
+          | None -> i)
+    in
+    aux 0
 
-let from_gen s = create (fill_buf_from_gen (fun id -> id) s)
+let from_gen gen = create (fill_buf_from_gen gen)
 
 let from_int_array a =
   let len = Array.length a in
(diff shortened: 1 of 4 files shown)


hhugo:
Preliminary benchmark seems to indicate that the performance profile is preserved for from_gen and from_channel and show a x6 speed improvement for `from_string`.

src/lib/sedlexing.ml
```diff
@@ -80,37 +78,45 @@ let set_position lexbuf position =
...
-        | Some c ->
-            buf.(pos + i) <- f c;
-            aux (i + 1)
-        | None -> i)
-  in
-  aux 0
+let refill_buf_from_gen gen =
+  let malformed = ref false in
+  fun buf pos len ->
+    if !malformed then raise MalFormed;
```
toots, line 84: Is this intended to let the lexer return something before raising `MalFormed`? In this case, is the value gonna be meaningful? Can the error be hidden in some cases e.g. if `refill_buf_from_gen` is not called again?

src/lib/sedlexing.ml
```diff
@@ -80,37 +78,45 @@ let set_position lexbuf position =
...
+    in
+    aux 0
 
-let from_gen s = create (fill_buf_from_gen (fun id -> id) s)
+let from_gen gen = create (refill_buf_from_gen gen)
 
 let from_int_array a =
+  let i = ref 0 in
+  (* copy the array to not be affected by future mutations *)
+  let a = Array.copy a in
```
toots, line 104: Does our API do mutations? If not, I would instead document that we are not expecting the array to be mutated while being used by the parser and consider those a user error. Copying the whole array will definitely have performance issues with large inputs.

src/lib/sedlexing.ml
```diff
@@ -80,37 +78,45 @@ let set_position lexbuf position =
...
-  let len = Array.length a in
-  {
-    empty_lexbuf with
-    buf = Array.init len (fun i -> a.(i));
-    len;
-    finished = true;
-  }
+let from_uchar_array buf =
+  (* copy the array to not be affected by future mutations *)
+  let buf = Array.copy buf in
```
toots, line 117: Same here.

src/lib/sedlexing.ml
```diff
@@ -266,47 +284,155 @@ module Chan = struct
...
-  let t = Chan.create ic (chunk_size * max_bytes_per_uchar) in
+module T_gen = struct
+  type t = {
+    b : Bytes.t;
+    gen : unit -> char option;
+    mutable len : int;
+    mutable pos : int;
+  }
+
+  let min_buffer_size = 64
```
toots, line 295: This seems arbitrary?

src/lib/sedlexing.ml
```diff
@@ -266,47 +284,155 @@ module Chan = struct
...
+       for the buffer. *)
+    if n <= 0 || n > min_buffer_size then invalid_arg "Sedlexing.Chan.ensure";
+    ensure_bytes_available t n
+
+  let get (t : t) i = Bytes.get t.b (t.pos + i)
+
+  let advance (t : t) n =
+    if t.pos + n > t.len then invalid_arg "advance";
+    t.pos <- t.pos + n;
+    assert (t.pos <= t.len)
```
toots, line 330: This seems redundant with the conditional above.

src/lib/sedlexing.ml
```diff
@@ -266,47 +284,155 @@ module Chan = struct
...
+
+  let available (t : t) = t.max - t.pos
+
+  let ensure_bytes_available (t : t) n =
+    let available = available t in
+    if available >= n then () else raise Missing_input
+
+  let ensure_bytes_available t ~can_refill:_ n =
+    (* [n] should not exceed the size of the buffer. Here we are
+       conservative and make sure it doesn't exceed the mininum size
```
toots, line 351: There's no minimum size for this module.

src/lib/sedlexing.ml
```diff
@@ -585,37 +613,45 @@ module Utf16 = struct
       aux apos len
   end
 
-  let from_gen s opt_bo = from_gen (Helper.from_gen opt_bo s)
+  let min_bytes_per_uchar = 2
+  let max_bytes_per_uchar = 4
```
toots, line 617: Constants with the same name are also defined above but with a different value for the min. Is that intended? I so, the name should be different to clarify the context of each constant. If not, they should be factored out.

src/lib/sedlexing.ml
```diff
@@ -369,62 +495,6 @@ module Utf8 = struct
...
-            let n3 = next_or_fail () in
-            let n4 = next_or_fail () in
-            if n2 lsr 6 != 0b10 || n3 lsr 6 != 0b10 || n4 lsr 6 != 0b10 then
-              raise MalFormed;
-            Uchar.of_int
-              (((n1 land 0x07) lsl 18)
-              lor ((n2 land 0x3f) lsl 12)
-              lor ((n3 land 0x3f) lsl 6)
-              lor (n4 land 0x3f))
-        | _ -> raise MalFormed
```
toots, line 405: I'm not sure where that logic is is in the new code.

Also, I believe these unicode-specific functions should be placed in a helper module outside of the generator code so they can be tested in isolation for strict adherence to the unicode standard..

toots (commented):
While I'm certainly happy to see some new code on the module, this seems more like a full rewrite than a quick cleanup.

I added some preliminary comment. Over all, the code could definitely be clarified and hopefully bring up some improvement, on performance, clarify, testing and also robustness. I like this idea of being able to raise `MalFormed` later, though I'm a little wary of the asynchronous nature of it.

Regarding the performances improvements on `from_string`, this is the same origin as https://github.com/ocaml-community/sedlex/pull/131 so I believe we should do a minor release with it to get the performances gain right away and pursue this as a major rewrite/release cycle.

src/lib/sedlexing.ml
hhugo: This is for utf16 (inside the utf16 module). the previous version is for utf8 (inside the utf8 module)

src/lib/sedlexing.ml
hhugo: This is no longer used, it was mostly a duplication of the function `next`, we now use that function `next` in all code path,

src/lib/sedlexing.ml
hhugo: The point of this change, extracted in #140, is to delay the malformed exception until we actually try to lex the malformed part of the input. 

> Can the error be hidden in some cases e.g. if refill_buf_from_gen is not called again?

If `refill` is not called again, it means that the lexer doesn't want to read the malformed part of the input, in which case it's expected to not raise Malformed.

PR #136, opened 2023-06-24
Format generator with ocamlformat

hhugo:


toots (approved):
Thanks! I have no idea why this file was left as generated. The original https://github.com/ocaml-community/sedlex/pull/78 PR has no clue either. I guess a left over from the original iterations there?

PR #137, opened 2023-06-24
from_int_array use from_gen

hhugo:
An invalid array no longer cause `Sedlexing.from_int_array` to raise an exception.


toots (approved):
Looking good, thanks!

PR #140, opened 2023-06-25
Delay MalFormed until really needed.

### LINKED ISSUE #133: lexbuf should handle invalid unicode input the same way regardless of how it is created.

hhugo:
replace #134
fix #133
When refilling the Uchar buffer of a lexbuf, we might encounter `MalFormed` input but the lexer might not try to read it in the end.
This PR makes sure we only raise `Malformed` when we actually need to read the problematic part of the input. The is done by raising `Malformed` only when we can't refill any uchar at all during a `refill` call.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 - Restore compatibility with OCaml 4.08
 - Use `Sedlexing.{Utf8,Utf16}.from_gen` to initialize UTF8 (resp. UTF16) lexing buffers from
   string.
+- Delay raising Malformed until actually reading the malformed part of the imput. (#140)
 
 # 3.1:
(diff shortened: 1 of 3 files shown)


toots (approved):
Looks great, thanks for the explanation.

toots:
I forgot to add: this probably could have a changelog entry!

hhugo:
I've added an entry

toots:
Thanks!

PR #141, opened 2023-06-26
Better unicode.ml generator

hhugo:
- no longer extract information from comments
- extract missing information from UnicodeData.txt
- track ignored properties so that one can codereview them
- make the parsing of Unicode data fail if lines don't have the expected format.

Based on #144


hhugo:
I've extracted some fix in #144

hhugo:
This is now based on top of #144.
Only the last commit is part of this PR.
once #144 is merged, this PR will only change how unicode.ml is generated but will no change its content.

src/generator/gen_unicode.ml
```diff
@@ -177,9 +145,58 @@ let () =
...
+      "po";
+      "ps";
+      "sc";
+      "sk";
+      "sm";
+      "so";
+      "zl";
+      "zp";
+      "zs";
+    ];
```
toots, line 180: I would make this a constant to make it clear what lists we're considering.

src/generator/gen_unicode.ml
```diff
@@ -177,9 +145,58 @@ let () =
...
+      "math";
+      "other_alphabetic";
+      "other_lowercase";
+      "other_math";
+      "other_uppercase";
+      "uppercase";
+      "white_space";
+      "xid_continue";
+      "xid_start";
+    ];
```
toots, line 200: Same

src/generator/gen_unicode.mli
```diff
@@ -0,0 +1 @@
+(* empty on purpose *)
```
toots, line 1: You can use `empty_module_interface_if_absent` from `dune` for that: https://dune.readthedocs.io/en/stable/dune-files.html#executable

toots (commented):
This looks great too. I'll have another pass once the parent PR is merged.

src/generator/gen_unicode.ml
hhugo: Are you proposing to lift the constant at toplevel ? I'm not sure what you have in mind that would make it clear (` what list we're considering`)

src/generator/gen_unicode.ml
hhugo: done

src/generator/gen_unicode.ml
hhugo: done

hhugo:
I've rebased this PR. ready for another round of review

toots (approved):
Looks great, thanks.

PR #147, opened 2023-06-27
Re-CI

hhugo:


hhugo:
@toots , Do you have any objection about removing `ci.yml` ?

PR #154, opened 2023-08-26
fix: use correct promotion mode

rgrinberg:
`(promote (until-clean))` is ignored when `--ignore-promoted-rules` is passed in later versions of dune.

See
https://github.com/ocaml/dune/issues/4401
https://github.com/ocaml/dune/pull/5956

### DIFF
--- a/src/syntax/dune
+++ b/src/syntax/dune
@@ -11,8 +11,7 @@
 
 (rule
  (targets unicode.ml)
- (mode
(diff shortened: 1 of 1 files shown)


rgrinberg:
In particular, this comment is relevant: https://github.com/ocaml/dune/issues/4401#issuecomment-808060902

toots:
Thanks!

PR #155, opened 2023-10-15
New functions to retrieve position

hhugo:
The goal is to improve the following pattern

```ocaml
let start, _ = Sedlexing.lexing_positions lb in
....
let _, stop = Sedlexing.lexing_positions lb in
```


toots (approved):
This looks good to me!

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# dev
+- Add API for retriving start and stop positions separatly (#155)
```
pmetzger, line 2: Typo. "separately".

pmetzger (commented):
I have a very trivial comment.

CHANGES.md
hhugo: and "retrieving"

toots:
Thanks for fixing the typos! I think that this is good to go!

PR #157, opened 2024-09-16
Add support for unicode 16.0.0.

toots:
This PR adds support for unicode `16.0.0`

### Notes:

* New non-binary properties were added to `DerivedCoreProperties.txt`. Those are not currently supported by the library and should be skipped for now.
* Regression tests report code point `0x1171e` missing in `mn`. This could be intentional.

~~There is quite a bit of noise due to some required module renaming to make the new old unicode ml file compile in the regression tests.~~

The only change otherwise is the addition, and use, of a `to_list` noop API to convert private `Cset` back into lists.

Otherwise, this is a fairly straight forward update.


pmetzger:
I haven't looked at Unicode 16; what are the non-binary properties for?

toots:
> I haven't looked at Unicode 16; what are the non-binary properties for?

I think that this documents it: https://www.gnu.org/software/libunistring/manual/html_node/Indic-conjunct-break.html

src/generator/gen_unicode.ml
```diff
@@ -46,7 +46,7 @@ let print_elements ch hashtbl cats =
           (fun (b, e) -> Printf.sprintf "0x%x, 0x%x" b e)
           (Cset.union_list (Hashtbl.find_all hashtbl c) :> (int * int) list)
       in
-      Printf.fprintf ch "  let %s = Sedlex_cset.of_list\n    [" c;
+      Printf.fprintf ch "  let %s = Sedlex_utils.Cset.of_list\n    [" c;
```
hhugo, line 49: I don't understand why we need this diff

examples/regressions.ml
```diff
@@ -32,6 +32,7 @@ let compare name (old_l : (int * int) list) (new_l : Sedlex_ppx.Sedlex_cset.t) =
     code_points
 
 let test new_l (name, old_l) =
+  let old_l = Sedlex_utils.Cset.to_list old_l in
```
hhugo, line 35: I think it would be better to manipulate cset directly. See #159

src/generator/gen_unicode.ml
```diff
@@ -91,7 +91,8 @@ let files =
               let interval = parse_interval interval in
               let prop = parse_prop prop in
               Hashtbl.add labels prop interval
-          | _ -> assert false );
+          (* TODO: support non-binary properties? *)
+          | _ -> () );
```
hhugo, line 95: I think it would be better to stay strict about what we accept. See #158.
That way, we always have to think about what to do whenever new format is introduced.

src/generator/gen_unicode.ml
toots: `unicode.ml` needs to compilable from within `src/syntax` and as `examples/unicode_old.ml`. 

Ideally, `unicode.ml` should be copiable to `unicode_old.ml` without any chances.

If you use `Sedlex_ppx.Sedlex_cset` is does not compile in `src/syntax`.

Happy to find a better solution but, also, I don't think that it matters.

examples/regressions.ml
toots: It still fails with `master` code (should be renamed `main` too). This call is just a pass-through. Looks like defining the type as `private` requires it. Again, not sure that this really matters..

toots:
```ocaml
type t = private (int * int) list

let to_list l = l
```

examples/regressions.ml
```diff
@@ -38,6 +38,7 @@ let compare name (old_ : CSet.t) (new_ : CSet.t) =
 let test new_l (name, old_l) =
   (* Cn is for unassigned code points, which are allowed to be
    * used in future version. *)
+  let old_l = CSet.to_list old_l in
```
hhugo, line 41: This line can be dropped, but you should also drop the lines below
```ocaml
let old_l =
      List.fold_left
        (fun acc (a, b) -> CSet.union acc (CSet.interval a b))
        CSet.empty old_l
    in
```

hhugo: Then to_list is no longer needed probably

toots: Addressed thanks!

examples/dune
```diff
@@ -1,6 +1,6 @@
 (executables
  (names tokenizer regressions complement subtraction repeat performance)
- (libraries sedlex sedlex_ppx)
+ (libraries sedlex sedlex_ppx sedlex_utils)
```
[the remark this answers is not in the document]
toots: Looks like the code doesn't need it.

src/generator/gen_unicode.ml
toots: Ok reverted.

examples/regressions.ml
toots: Actually this has been put back in place. I haven't looked at this code in a while and, at face value, the `to_list` call is just to work around a change in type declaration while removing the fold definitely is changing the code's runtime semantics so the safter choice seems to be to keep the new `to_list`.

toots:
Ok I have cleaned up this PR and minimized the diff. I think that it is ready for merge now.

pmetzger:
There seem to be conflicts that would prevent a merge?

toots:
> There seem to be conflicts that would prevent a merge?

You mean `git` conflicts?

hhugo (approved):
Looks good to me.

PR #160, opened 2024-10-20
Prepare for 5.2 AST bump

patricoferris:
This PR is a patch for an upcoming release of ppxlib where the internal AST is [bumped from 4.14 to 5.2](https://github.com/ocaml-ppx/ppxlib/pull/514). Until that is merged and released, there is no reason to merge this PR.

To test these changes, you can use the following opam-based workflow. I've made releases of most of these patches to an opam-repository overlay.

```shell
opam switch create ppxlib-bump --repos=ppxlib=git+https://github.com/patricoferris/opam-repository#5.2-ast-bump
opam install <your-package>
```

The following describes the most notable changes to the AST.

### Functions
---

#### Currently 

In the parsetree currently, functions like:

```ocaml
fun x y z -> ...
```

Are represented roughly as

```ocaml
Pexp_fun(x, Pexp_fun (y, Pexp_fun(z, ...)))
```

Functions like:

```ocaml
function A -> ... | B -> ...
```

Are represented roughly as

```ocaml
Pexp_function ([ case A; case B ])
```

#### Since 5.2

All of these functions now map to a single AST node `Pexp_function` (note, this is the same name as the old cases function). The first argument is a list of parameters meaning:

```ocaml
fun x y z -> ...
```

Now looks like:

```ocaml
Pexp_function([x; y; z], _constraint, body)
```

And the `body` is where we can either have _more_ expressions (`Pfunction_body _`) or cases (`Pfunction_cases _`). That means:

```ocaml
function A -> ... | B -> ...
```

Has an empty list of parameters:

```ocaml
Pexp_function([], _, Pfunction_cases ([case A; case B]))
```

### Local Module Opens for Types

Another feature added in 5.2 was the ability to locally open modules in type definitions. 

```ocaml
module M = struct
  type t = A | B | C
end

type t = Local_open_coming of M.(t)
```

This has a `Ptyp_open (module_identifier, core_type)` AST node. Just like normal module opens this does create some syntactic ambiguity about where things come from inside the parentheses. The approach of these patches is to locally open the same module in any code that is generated.


toots:
Thanks!

smorimoto:
It would be nice to merge this to release the next release!

toots:
Looks like `ppx_expect` also needs a release.

toots:
We'd have to make it clear that we only support OCaml above `5.2` after this is merged. This is a bit annoying but tis what it is.

patricoferris:
Thanks for taking a look @toots !

Sorry if the original message wasn't clear. Whilst we are bumping the _internal_ AST of ppxlib to 5.2, all of the migrations are still in place so that ppxes will support 4.08 and up :))

hhugo:
We don't need `pexp_function_cases`.
Here is an alternative compatible with multiple version of ppxlib
```diff
diff --git a/src/syntax/ppx_sedlex.ml b/src/syntax/ppx_sedlex.ml
index 601ac94..f37a3bd 100644
--- a/src/syntax/ppx_sedlex.ml
+++ b/src/syntax/ppx_sedlex.ml
@@ -4,6 +4,7 @@
 
 open Ppxlib
 open Ast_builder.Default
+open Ast_helper
 
 (* let ocaml_version = Versions.ocaml_408 *)
 
@@ -232,7 +233,7 @@ let gen_state (lexbuf_name, lexbuf) auto i (trans, final) =
     [
       value_binding ~loc
         ~pat:(pvar ~loc (state_fun i))
-        ~expr:(pexp_function ~loc [case ~lhs ~guard:None ~rhs:body]);
+        ~expr:(Exp.fun_ ~loc Nolabel None lhs body);
     ]
   in
   match best_final final with
```

Issue #162, opened 2024-11-01
CI: re-enable the build workflow somehow 


hhugo:
The build workflow doesn't seem to run anymore. Can someone re-enable it ?

> This scheduled workflow is disabled because there hasn't been activity in this repository for at least 60 days. Enable this workflow to resume scheduled runs.

PR #164, opened 2024-11-01
Update ci

hhugo:


hhugo:
One need to fix https://github.com/ocaml-community/sedlex/issues/162

toots:
This looks good but it seems that you need this to be in the main repo to be executed. Will merge and follow-up

Issue #167, opened 2025-05-18
UTF-8 decoder accepts overlong sequences (non-shortest form), should be updated to at least Unicode 3.1


edwintorok:
Recent versions of Unicode forbid overlong UTF-8 sequences. Looking through Unicode version history that requirement has been introduced here: https://www.unicode.org/versions/corrigendum1.html

Currently sedlex appears to have a Unicode 3.0 UTF-8 decoder, which accepts certain sequences that should be rejected.

Here is a file to exercise the various corner cases: https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt
A small .ml file that acts as an 'echo':
```
let b = Buffer.create 4

let rec token buf =
      match%sedlex buf with
        | eof -> print_endline ""
        | any ->
            Buffer.clear b;
            Buffer.add_utf_8_uchar b (Sedlexing.lexeme_char buf 0);
            output_string stdout (Buffer.contents b);
            token buf
        | _ -> assert false

let get_line () = In_channel.input_line stdin

let () =
  get_line |> Seq.of_dispenser |> Seq.iter @@ fun line ->
  let lexbuf = Sedlexing.Utf8.from_string line in
  try
    token lexbuf
  with Sedlexing.MalFormed ->
    print_endline "MALFORMED"
```

Running it on the above file it fails a few tests (note that the above file was written with an even older version of Unicode, and doesn't require rejecting > 0x10FFF):
```
With a safe UTF-8 decoder, all of the following five overlong                 |
representations of the ASCII character slash ("/") should be rejected         |
like a malformed UTF-8 sequence, for instance by substituting it with         |
a replacement character. If you see a slash below, you do not have a          |
safe UTF-8 decoder!                                                           |
                                                                              |
4.1.1 U+002F = c0 af             = "/"                                        |
4.1.2 U+002F = e0 80 af          = "/"                                        |
4.1.3 U+002F = f0 80 80 af       = "/"                                        |

4.2  Maximum overlong sequences                                               |
                                                                              |
Below you see the highest Unicode value that is still resulting in an         |
overlong sequence if represented with the given number of bytes. This         |
is a boundary test for safe UTF-8 decoders. All five characters should        |
be rejected like malformed UTF-8 sequences.                                   |
                                                                              |
4.2.1  U-0000007F = c1 bf             = ""                                   |
4.2.2  U-000007FF = e0 9f bf          = "߿"                                   |
[...]

PR #168, opened 2025-05-19
Implement Corrigendum #1: UTF-8 Shortest Form

toots:
Fixes: #167


hhugo:
Should we declare test/stress_test.ml as text in `.gitattributes` ?

hhugo:
Is it really necessary to add the ppx_blob test dependency ? Can't you just use `open_in`,`in_channel_length` and `really_input_string` from the stdlib ?

toots:
> Should we declare test/stress_test.ml as text in `.gitattributes` ?

Looks like `github` might have its own opinion on it but it's marked as `text` and `diff` (to make sure it also shows up in diffs).

hhugo:
> > Is it really necessary to add the ppx_blob test dependency ? Can't you just use `open_in`,`in_channel_length` and `really_input_string` from the stdlib ?
> 
> We can. `dune` doesn't seem to have any directive to force a file to be copied to `_build` before building except for the `ppx_preprocessor` directive. Once I started using it, it made more sense to actually use a pre-processor to include the file.

It does, it was implemented in #169

toots:
If you feel strongly that it should be done this way feel free to re-send a PR, I won't oppose it.

PR #179, opened 2026-03-11
Tag delay optimization

hhugo:
## Summary

- When a `Set_position` tag fires on every transition entering a cycle state, remove it from those transitions and emit `Set_position` with offset 1 (`pos - 1`) on exit transitions. Turns O(n) tag writes per loop into O(1) on exit.
- Handles multi-state cycles via SCC analysis (Tarjan's algorithm) and multi-exit cycles: exits from other cycle states are safe when they only reach rules that don't read the tag.
- `Set_position { cell; offset }` with offset 0 for current position and offset 1 for previous code point (subtracted from pos).

## Test plan

- [x] Codegen expect tests (DOT graphs + generated code)
- [x] Runtime tests for multi-state cycles, multi-exit cycles, backtracking
- [x] `test_realistic.ml` multi-rule lexer baseline

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # unreleased
 - Support nested `let..in` for `[%sedlex.regexp?]` definitions
-- Add support for named captured group (#177, #178)
+- Add support for named captured group (#177, #178, #179)
 
 # 3.7 (2025-10-06)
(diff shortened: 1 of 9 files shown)


hhugo:
replaced by #209

PR #193, opened 2026-04-01
Defer tag writes past fixed-length neighbors

hhugo:
## Summary

- Prefer end tag over start tag for fixed-length element optimization — delays the tag write past the element
- Allocate boundary tags for fixed-length tuple elements without a right-position anchor, so that as-binding tags fire as late as possible in the automaton
- Dead tag elimination pass (`Sedlex.optimize`) strips unused boundary tags and remaps live ones to a dense range
- Inner tuples communicate boundary anchors to enclosing aliases via `aux`'s return triple

## Test plan
- [ ] Existing expect tests pass (codegen, realistic, basic)
- [ ] New test: deferred start tag past inner prefix (`("0x", Plus hexa) as x`)
- [ ] Verify dead boundary tags are eliminated (mem_cells unchanged for patterns without aliases)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # unreleased
 - Support nested `let..in` for `[%sedlex.regexp?]` definitions
 - Add support for named captured group (#177, #178)
+- Optimize tag placement: defer tag writes past fixed-length neighbors (#193)
 
 # 3.7 (2025-10-06)
 - Update to unicode 17.0.0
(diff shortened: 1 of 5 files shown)


pmetzger:
I am unlikely to be able to meaningfully review this, but I will happily do the merge if people tell me to.

hhugo:
rebased on top of #196

pmetzger:
@hhugo I added you to the writers for this repository, so you should be able to update such things when you believe it is appropriate. Please be careful about testing and correctness, especially with machine generated code.

pmetzger:
@hhugo Just to repeat: when you feel this is good enough, you can ask for consensus, and if no one answers, you should feel free to commit.

PR #194, opened 2026-04-02
add NFA/DFA oracle tests

hhugo:
## Summary

- **Extract `sedlex.compiler` library**: Move `Sedlex`, `Cset` (formerly `Sedlex_cset`) out of the PPX into a standalone `sedlex.compiler` library. Absorb `sedlex.utils` into it. Non-PPX code (tests, tools) can now depend on the regexp/DFA engine without pulling in ppxlib.
- **Add `Cset.mem`**: Code point membership test for character sets.
- **Add NFA/DFA oracle testing**: A `sedlex_oracle` library with a per-thread NFA simulator that serves as ground truth for the DFA. Compares acceptance and tag/binding values between NFA and DFA, printing `FIXME` on mismatches. Includes hand-written expect tests and QCheck property-based tests. Documents the known FIXME in `add_node` where the shared tag-op list causes wrong captures for overlapping Star/Plus + as-binding patterns.

## Test plan

- [x] `dune build` succeeds
- [x] `dune runtest` passes (all existing tests + new oracle tests)
- [x] QCheck acceptance tests (2000 random regexps) pass
- [x] QCheck tag-aware tests find and report known DFA limitations with `FIXME` prefix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #196, opened 2026-04-07
Add IR to compiler; validate in smart constructors

hhugo:
## Summary

- Introduce `Ir.t`, a pure intermediate representation for sedlex patterns that captures regexp structure and named captures **before** tag allocation. Lives in `src/compiler/ir.ml` with no ppxlib dependency.
- Add `Sedlex.compile_ir : Ir.t array -> compiled_ir` to the compiler library. It handles tag allocation (Start_plus/End_minus optimizations), discriminator insertion for or-patterns, and DFA construction.
- The PPX becomes a thin translator from OCaml pattern AST to `Ir.t` via `ir_of_pattern`, delegating all optimization decisions to the compiler.
- Flatten `Alt` from binary to n-ary (`Alt of t list`), simplifying discriminator assignment to a single pass and removing the `reusable_cell` heuristic.
- Move structural validation into IR smart constructors returning `(t, string) result`: `capture` checks shadowing, `alt` checks name consistency across branches, `star`/`plus`/`rep` reject inner captures. The PPX uses a local `unwrap` helper to bridge results to ppxlib error reporting.
- Use `SSet` instead of `string list` for `capture_names`.
- Improve error messages: bare operators (Star, Plus, Opt, etc.) without arguments now produce specific errors instead of falling through to "not a valid regexp"; unknown constructors produce "unknown sedlex operator X".

## Motivation

The PPX previously interleaved pattern parsing with tag allocation decisions (`bind` vs `bind_start_only` vs `bind_end_only`), discriminator handling, and `fixed_length` computation. This made it impossible to test the compilation pipeline without ppxlib, and spread compiler concerns across two packages.

With the IR, tests can construct `Ir.t` values directly and exercise the full `compile_ir` path. Future optimizations (tag delay, register allocation) can be added to the compiler without touching the PPX.

## Test plan

- [x] `dune build` succeeds
- [x] `dune runtest` passes — all existing expect tests produce identical output (modulo disc cell renumbering from the flat Alt change)
- [x] Error messages preserved for all validation cases (captures inside Star/Plus/Rep/Opt/Compl/Sub/Intersect, mismatched or-pattern names)

### DIFF
--- a/src/compiler/ir.ml
+++ b/src/compiler/ir.ml
@@ -61,16 +61,16 @@ let rec fixed_length = function
       else None
   | Star _ | Plus _ -> None
 
+module SSet = Set.Make (String)
+
 let rec capture_names_acc acc = function
   | Chars _ | Eps -> acc
-  | Capture (name, inner) ->
-      let acc = if List.mem name acc then acc else name :: acc in
-      capture_names_acc acc inner
+  | Capture (name, inner) -> capture_names_acc (SSet.add name acc) inner
   | Seq elems -> List.fold_left capture_names_acc acc elems
   | Alt branches -> List.fold_left capture_names_acc acc branches
   | Star inner | Plus inner | Rep (inner, _, _) -> capture_names_acc acc inner
 
-let capture_names t = capture_names_acc [] t |> List.sort_uniq String.compare
+let capture_names t = capture_names_acc SSet.empty t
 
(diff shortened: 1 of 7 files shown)


hhugo:
@toots, I think you'll like this PR as well. It will help with #194 quite a bit.

PR #199, opened 2026-06-11
Rewrite determinization as a Laurikari TDFA for correct capture semantics

hhugo:
## Summary

This series fixes incorrect submatch positions for `as` bindings and puts the capture machinery on a proper theoretical footing.

**The bug.** The previous determinization attached one flat tag list per transition. A `Star`/`Plus` loop whose epsilon closure contained a capture's start-tag node re-fired the position write on every iteration, clobbering the position recorded by the path already inside the capture — yielding wrong, empty, or garbage submatches (e.g. `Star 'a', (('a', Plus 'b') as x)`).

**The fix.** Determinization is now a proper Laurikari tagged-DFA construction, modeled on ocamllex's `lex/lexgen.ml`:

- A DFA state is an ordered list of configurations (NFA node, per-tag register map); order is priority. Epsilon closure is a DFS in `eps`-list order and the first path to a node wins, giving leftmost-greedy disambiguation among maximal-length matches.
- State lookup is modulo bijective register renaming. Hitting an existing state emits Set/Copy register moves as a **parallel move**; the PPX let-binds clobbered Copy sources, so no move ordering or temporary cells are needed (unlike ocamllex's interpreted C engine).
- A final rename pass collapses **conflict-free tags** (never two distinct registers in one DFA state) into their canonical cell and drops the resulting no-op copies. Most patterns end up with exactly one cell per logical tag; only genuinely conflicted shapes (e.g. a loop before an overlapping capture) pay for extra working registers.

**Verification.** A new oracle harness (`test/oracle/`) compares a brute-force leftmost-greedy reference matcher over `Ir.t` against a faithful interpreter of the compiled DFA (including mark/backtrack and final-ops semantics) on exhaustive small inputs plus deterministic seeded random sweeps. The codegen expect-tests now also include non-eliminable counterparts for every `as`-binding shape whose positions would otherwise be computed statically, so the generated tag operations are visible in the expected output.

## Commits

- `Add Cset.mem for code point membership testing`
- `Add oracle tests: reference matcher vs compiled tagged DFA`
- `Add Copy tag operation to runtime, PPX, and oracle interpreter`
- `Rewrite determinization as a proper Laurikari TDFA`
- `Compile register moves as parallel moves, dropping temp cells`
- `Write conflict-free tags directly to their canonical cell`
- `Simplify Seq lowering to two list folds`
- `Add non-eliminable counterparts for static as-binding tests`
- `Remove duplicate or-pattern as-binding test`

## Future work (tracked as Goal notes in test/codegen/test_gen.ml)

- Cross-rule cell sharing (non-interfering rules reusing cells)
- Dead tag elimination on shared prefixes
- Self-loop tag delay (`set_mem_prev` on exit instead of O(n) writes)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 # unreleased
 - Support nested `let..in` for `[%sedlex.regexp?]` definitions
 - Add support for named captured group (#177, #178)
+- Rewrite determinization as a Laurikari tagged DFA, fixing incorrect
+  submatch positions when a repetition loop overlaps a capture (#199)
 
 # 3.7 (2025-10-06)
 - Update to unicode 17.0.0
--- a/dune-project
+++ b/dune-project
@@ -27,4 +27,5 @@ extension.")
    (ppxlib (>= 0.26.0))
    gen
    (ppx_expect :with-test)
+   (qcheck-core :with-test)
    (menhir :with-test)))
--- a/sedlex.opam
+++ b/sedlex.opam
@@ -22,6 +22,7 @@ depends: [
   "ppxlib" {>= "0.26.0"}
   "gen"
   "ppx_expect" {with-test}
+  "qcheck-core" {with-test}
   "menhir" {with-test}
   "odoc" {with-doc}
 ]
--- a/src/compiler/cset.ml
+++ b/src/compiler/cset.ml
@@ -74,5 +74,10 @@ let complement c =
   in
   match c with (-1, j) :: l -> aux (succ j) l | l -> aux (-1) l
 
+let rec mem c = function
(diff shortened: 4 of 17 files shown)


pmetzger:
If you're interested in the theory here, a couple of important papers are associated with the re2c effort, see: https://re2c.org/#papers

pmetzger:
@hhugo What's your intent for this?

hhugo:
> @hhugo What's your intent for this?

converted to a draft while I finish my investigation. Sorry for the noise

pmetzger:
Don't be sorry! It's good that you're doing the work!

PR #203, opened 2026-09-18
Plumbing for the TDFA rewrite: Copy op, inline records, accept field, parallel moves

hhugo:
Preparation for the tagged-DFA rewrite that fixes #199 (series: #200, #201, #202). Four commits that add everything the rewrite needs outside the compiler's determinization, without the compiler producing any of it yet. **Generated code and every expect output are byte-identical** (`dune runtest` prints no diff after each commit); the review check is that nothing new is emitted.

Commits:

1. **`Copy` tag operation.** Copies the raw contents of one memory cell into another, preserving the position/value encoding. Runtime `__private__copy_mem`, PPX generation, DOT label `tN<-tM`. Not emitted yet.
2. **Inline records for `tag_op`.** `Set_position {dst}`, `Set_value {dst; value}`, `Copy {dst; src}`: the rewrite manipulates these in allocation, register moves and renaming, where named fields read better than tuple positions.
3. **`accept : accept option`, with `accept = { rule; final_ops }`, replaces `finals : bool array` on DFA states.** Rule priority is resolved in the compiler (`accepting_rule`) instead of the PPX's `best_final`, and the accepting state carries the operations to run on entry just before `Sedlexing.mark`. The rewrite will use them to copy the accepting path's working registers into the canonical cells read by binding extraction; the list is `[]` for now, so `gen_tag_ops` wraps nothing. Nothing driving the automaton needs a state's other accepting rules, so `finals` goes away and the DOT printer labels an accepting state with the winning rule only; no expect output listed more than one rule, so the graphs are unchanged.
4. **Parallel-move `gen_tag_ops`.** A transition's operations form a parallel move: a `Copy` reads its source as it was before the list ran. Sources the list also writes are saved in let-bound locals through the new raw accessors `__private__mem_get` / `__private__mem_set`; `op_dest` is exported for it. No `Copy` is emitted yet.

With these in, the rewrite's diff outside `src/compiler/sedlex.ml` is confined to expect outputs (plus the matching one-line changes in the #202 interpreter).


pmetzger:
Very cool!

PR #205, opened 2026-09-18
Add support for unicode 18.0.0

hhugo:
Update the Unicode tables from 17.0.0 to 18.0.0, following the procedure in `src/generator/README.md`.

### Commits

1. **Fix regressions allowlist in unicode regression test.** `regressions` in `examples/regressions.ml` lists code points of the old set that are expected to be missing in the new one. Since #159 the validity checks were inverted: entries present in the old set were reported as invalid and dropped, so the allowlist could never suppress a reported diff. An entry is now valid when it is in the old set and absent from the new one.
2. **Add support for unicode 18.0.0.** Bump `base_url`, regenerate `src/syntax/unicode.ml`, move the 17.0.0 table to `examples/unicode_old.ml`, update `test_versions`, add a CHANGES entry.

### Notes

- No regression between 17.0.0 and 18.0.0: every exported category and property of 17.0.0 is a subset of its 18.0.0 counterpart, so `regressions` stays empty.
- No new binary property was added to `PropList.txt` or `DerivedCoreProperties.txt` since 16.0.0, so the exported list is unchanged.
- For the record, the 16.0.0 → 17.0.0 update (#172) had one unrecorded regression: U+0295 LATIN LETTER PHARYNGEAL VOICED FRICATIVE changed `General_Category` from `Ll` to `Lo`, so it left `ll` and `lowercase`. The test printed it but does not fail on it. It is no longer visible now that the old table is 17.0.0.


toots (approved):
Thanks!

PR #208, opened 2026-09-20
Stop copying the capture cells in Sedlexing.mark and backtrack

hhugo:
Stops copying the capture cells in `Sedlexing.mark` / `Sedlexing.backtrack`. **Stacked on #204** (base: the `pr4-tdfa-rewrite` branch), one commit. #209 and #207 are stacked on it.

**Cost today.** Since #177, `mark` copies every memory cell of the lexbuf into a second array on each entry into an accepting state, and `backtrack` copies them back. Marking is frequent (every character of an identifier is an accepting state), and the array only grows, to the largest `init_mem` requested on that lexbuf: once one `match%sedlex` block with captures has run, every block pays the copy on every mark, captures or not.

**Why it is not needed after #204.** The cells a rule's action reads, its canonical cells, are only written by the final operations of that rule's accepting states, just before `mark`; transitions write working registers. When a longer match fails, the canonical cells of the marked rule still hold the values of the last accepting state.

#207, which lets unambiguous tags live in their canonical cell, carries the matching condition: a tag that could be written after an accepting state of its rule, on the way to a failure, keeps a working register. So collapsing there does not rely on the copy.

**Change.** `__private__mem_saved` is gone, `mark`/`backtrack` no longer touch the cells, `refill` shifts one array. New test `fallback_keeps_submatches`: patterns that accept, go on with a longer alternative recording the same sub-match again, then fail.

**Verification.** The #202 oracle, run locally with the copy removed from its interpreter too: no disagreement over 300k random cases at this tip.


hhugo:
Waiting for #204 to be merged first.

